PR #84, opened 2020-04-07
Prevent creation of contacts with identical account numbers

### LINKED ISSUE #82: Should not be able to create identical contact/external accounts
Prevent users from being able to create contacts or external accounts with identical account numbers & routing numbers.

Instead, show an error message:  e.g. "contact already exists"

djmailhot:
Resolves #82 

When a user attempts to create a new contact that already exists, the frontend will fail the transaction and display the following error message:  "[Payment|Deposit] failed: account already exists as a contact"


DanSanche:
Should we also prevent assigning the same name to multiple contacts?

src/frontend/flask_server.py
```diff
@@ -146,21 +146,26 @@ def payment():
...
-                             LOCAL_ROUTING,
-                             False)
+                response = _add_contact(label,
+                                        recipient,
+                                        LOCAL_ROUTING,
+                                        False)
+                if response.status_code == 409:
+                    # contact not allowed.
+                    msg = 'Payment failed: ' + response.json()['msg']
+                    return redirect(url_for('home', msg=msg))
```
DanSanche, line 156: - It might be better to throw an exception in _add_contact and catch it with the others
- Should we only stop the transaction if _add_contact returns 409? What if the contact service throws a different error? I'd say we should treat those cases the same

djmailhot:
> Should we also prevent assigning the same name to multiple contacts?

Good point.  We can also do that.

src/frontend/flask_server.py
djmailhot: Good idea!  Both points done.

PR #90, opened 2020-04-08
Python improvements

DanSanche:
tried to make the python services more production-ready:
- use `gcr.io/google-appengine/python` as a base image
- pin all dependencies to a specific version
  - use requirements.in and pip-compile to pin cascading dependencies
- use gunicorn as server instead of Flask's built in server (which isn't recommended for production)
- switched out Flask's logger to log to gunicorn
- use `APP.config` to store configuration variables


src/contacts/Dockerfile
```diff
@@ -12,18 +12,30 @@
...
+
 # show python logs as they occur
 ENV PYTHONUNBUFFERED=0
 
-# get packages
-WORKDIR /contacts
-COPY requirements.txt .
-RUN pip install --trusted-host pypi.python.org -r requirements.txt
+# Create a virtualenv for the application dependencies.
+RUN virtualenv -p python3.7 /env
```
bharathkkb, line 25: Is there a particular reason why we are using venv, usually I have only seen this in the context of local dev.
Or are we evaluating multi stage builds for leaner images?

DanSanche: I'm following the pattern [recommended in the repo](https://github.com/GoogleCloudPlatform/python-runtime#kubernetes-engine--other-docker-hosts) for the official Google python image

I'd guess the reason is to make sure that all dependencies are captured in requirements.txt and we're not accidently relying on the system packages for anything, as those may shift over time

bharathkkb: oh makes sense

Issue #53, opened 2020-03-25
Add end-to-end tests to CI


murog:
- [x] Add new e2e step to actions workflow (#92) 
- [x] 1: User can log in with default account and see correct balance
- [x] 2: User is blocked from signing in with bad credentials
- [x] 3: User can create account and see zero balance
- [ ] 4: User can deposit funds
    * See balance update
    * See transaction in history
    * See new contact show up
- [ ] 5: User can transfer funds
    * See balance update
    * See transaction in history
    * See new contact show up
- [ ] 6: User is blocked from sending invalid data
    * Amount > balance
    * Amount <= 0
    * Amount with more than 2 decimal digits
    * Badly formatted routing or account numbers
- [x] 7: User is redirected from /home to /login when not authenticated
- [x] 8: User is redirected from /login and /signup to /home when already authenticated

murog:
Fixed in #183

PR #92, opened 2020-04-09
Initialize Cypress in project and CI

murog:
Initializes cypress for issue #53 
- [x] adds e2e directory with 
    * cypress, simple test and instructions
    * cypress is configured with screenshots disabled, opened #93 
- [x] add cypress to CI
    * port-forwards frontend to localhost:3000
    * pings localhost:3000 to ensure connection
    * runs cypress tests using [cypress docker image](https://github.com/cypress-io/cypress-docker-images/tree/master/included)


.github/workflows/smoke-tests.yml
```diff
@@ -87,4 +87,18 @@ jobs:
...
-
+    - name: Port Forward frontend
+      run: |  
+        set -x
+        port_forward() {
+            # get name of frontend pod
+            local frontend_pod=$(kubectl get pods -l=app=frontend -o=jsonpath='{.items[0].metadata.name}')
+            kubectl port-forward $frontend_pod 3000:8080   
+        }
+        port_forward &
```
DanSanche, line 98: Do we know if this process is cleaned up after the CI run?

.github/workflows/smoke-tests.yml
```diff
@@ -87,4 +87,18 @@ jobs:
...
+    - name: Ping frontend
+      run: |
+          curl --silent --output /dev/null --show-error --fail localhost:3000 
+    - name: UI tests
+      run: |
+        docker run -t -v $PWD/e2e:/e2e -w /e2e --network=host cypress/included:4.3.0
```
DanSanche, line 104: If cypress is already packaged as a container, can't we just 'kubectl run' it in the cluster? That way we wouldn't have to worry about port forwarding. We'd just need to find a way to mount the e2e folder

.github/workflows/smoke-tests.yml
murog: On my fork the Complete Job logs
```
Cleaning up orphan processes
Terminate orphan process: pid (15260) (bash)
Terminate orphan process: pid (15272) (kubectl)
```
And I've been able to verify on my fork's runner that it's no longer running.  Will verify on this repo's runner.

.github/workflows/smoke-tests.yml
murog: I would have to dig in a bit to verify that change on my end.  Any reasons why the port-forwarding should be avoided?

DanSanche: - I'm concerned that the port-forward process could stick around between runs, reserving port 3000 and causing issues (although from your last comment I'm not sure if we need to worry about this)
- Kubernetes and Docker have their own abstractions for stuff like networking, and it would be a lot cleaner if we could stick to one platform instead of trying to get them to talk. I'm assuming this test will get more complex over time, so we might appreciate being in the same virtual network as all the other services at some point later.

But overall if the Github Actions docs say it's ok to run background processes and they will be cleaned up for us, I'm not too worries either way

e2e/cypress/support/index.js
```diff
@@ -0,0 +1,40 @@
...
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands'
+
+// Alternatively you can use CommonJS syntax:
+// require('./commands')
```
DanSanche, line 36: nit: you may want to remove some of these boilerplate comments

DanSanche:
I haven't been able to find docs for GitHub Actions talking about background tasks, but a GitHub engineer seems to [recommend it here](https://github.community/t5/GitHub-Actions/How-do-I-run-commands-concurrently/td-p/36057), so I think it should be a safe practice. It may still be worth researching running tests in Kubernetes to see if it's feasible, but it looks like either option can work, so LGTM

My only other suggestion would be to consider moving the e2e folder into either:
- a common `tests/` folder if that's where we'll keep unit tests
- the `.github/workflows` folder if unit tests will live alongside the code for each service (since this is effectively our tests folder)

murog:
@DanSanche Thanks for looking into this! Will keep docker and port-forwarding for now.  Removed `e2e/` dir and moved cypress files into `.github/workflows/ui-tests`

.github/workflows/smoke-tests.yml
bharathkkb: As an added precaution, can we not add a step to clean up any process on particular ports if that is a concern?

murog: We can definitely add a clean up process if it becomes an issue.  I don't think it's necessary to add at the moment since the actions runner [cleans up orphaned processes](https://github.com/actions/runner/blob/b0a1294ef5751746eb2b4815c4e3e8a6d575833c/src/Runner.Worker/JobExtension.cs#L271) and surfaces it in the [check's logs](https://pipelines.actions.githubusercontent.com/BR7w0sd8lc5VMu0meoJMolOSMdu4pnfw1Nj5iUBFLi2lDnme12/_apis/pipelines/1/runs/440/signedlogcontent/3?urlExpires=2020-04-13T20%3A01%3A35.4466587Z&urlSigningMethod=HMACV1&urlSignature=QPTqk66ghy3gC4C81h5aoGY4OyzWQL2Igu3oO3vqNeo%3D)

The runner's logs also show this:
```
[2020-04-10 22:16:31Z INFO JobExtension] Inspecting process environment variables. PID: 4440 (kubectl)
```

PR #95, opened 2020-04-09
Move python database related ops to a shared module

bharathkkb:
Changelist

- Refactor contacts and user services to move database related private methods to a new module
- Set build context for python services at src level so that we can copy pylibs
- Symlink shared pylibs to contacts/user services for ease of local development


src/pylibs/db/sql_adapter.py
```diff
@@ -0,0 +1,158 @@
...
+            'contacts',
+            MetaData(self.engine),
+            Column('username', String),
+            Column('label', String),
+            Column('account_num', String),
+            Column('routing_num', String),
+            Column('is_external', Boolean),
+        )
+        # open a connection
+        self.conn = self.engine.connect()
```
bharathkkb, line 54: @djmailhot @DanSanche not sure if opening a connection and not closing it until end of pod lifecycle is a best practice. I dont have alot of SQL experience, but from what I found online it seems we should be atomically opening and closing connections per operation.

djmailhot: That's a good point!

SQL has limits to the number of concurrent connections (PostgreSQL defaults to 100), so opening/closing connections on each request would 1) limit the maximum QPS of the service, and 2) add overhead latency to each request.

HOWEVER, it appears that sqlalchemy automatically employs [connection pooling](https://docs.sqlalchemy.org/en/13/core/pooling.html), which means **we can and should indeed open/close connections on each request**.

src/pylibs/db/sql_adapter.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def add_user(self, user):
```
djmailhot, line 56: (I know you just copied the existing code, so apologies for the following.)

To be clear, as the code is written, this `user` parameter is actually a collection of attributes from an HTML form, including fields like `password` and `password-repeat`.

It also means that `accountid` needs to be added and `password` needs to be hashed into `passhash` before this object starts looking like an actual `user` object.  This sql adaptor class is meant to be a thin database wrapper and should probably not need to responsible for all that.

Suggestion:
(strong) move the password hashing back into `userservice.py`
(maybe) move accountid generation back into `userservice.py`
- would likely require implementing `sql_adapter.get_user(accountid)`
- This one is a little questionable and I could see the case for leaving it here in the adaptor because SQL is often left to generate object ids with `GENERATED AS IDENTITY`.  (We do it manually so that our user accounts don't look like `1000000001`, `1000000002`, `1000000003`, etc.)

src/pylibs/db/sql_adapter.py
bharathkkb: aha cool, I will make that change

bharathkkb: Done

src/pylibs/db/sql_adapter.py
bharathkkb: I have moved password hashing back into `userservice.py`
I havent moved account id generation back to `userservice.py` yet although I okay with implementing `sql_adapter.get_user(accountid)` and doing this. 
@Daniel-Sanche any thoughts?

src/pylibs/db/tests/test_db_sql.py
```diff
@@ -0,0 +1,84 @@
...
+        user['username']='bar'
+        user['accountid']=self.DB.generate_accountid()
+        #add bar_user to db
+        self.DB.add_user(user)
+        #try to add same user again
+        self.assertRaises(IntegrityError,self.DB.add_user,user)
+
+    def test_get_user(self):
+        user=EXAMPLE_USER.copy()
+        #create a user with username bar
```
djmailhot, line 46: bar -> baz

src/pylibs/db/tests/test_db_sql.py
```diff
@@ -0,0 +1,84 @@
...
+    def setUpClass(cls):
+        cls.DB = DatabaseHelper('SQL', 'sqlite:///:memory:').database
+        cls.DB.users_table.create(cls.DB.engine)
+        cls.DB.contacts_table.create(cls.DB.engine)
+
+    def test_add_user_foo(self):
+        user=EXAMPLE_USER.copy()
+        #create a user with username foo
+        user['username']='foo'
+        user['accountid']=self.DB.generate_accountid()
```
djmailhot, line 31: (med) Suggest creating separate tests for generate_accountid()

In this add_user() test, you end up also testing generate_accountid().  You should probably separate the tests by simply passing a literal accountid.

In tests for generate_accountid(), you can look into mocking the values for random.randint and verifying that generate_accountid() actually avoids duplicate account ids.

src/pylibs/db/sql_adapter.py
```diff
@@ -0,0 +1,130 @@
...
+                statement = self.users_table.select().where(self.users_table.c.accountid == accountid)
+                logging.debug('QUERY: %s', str(statement))
+                result = conn.execute(statement).first()
+                logging.debug('RESULT: %s', str(result))
+                # If there already exists an account, try again.
+                if result is not None:
+                    accountid = None
+        return accountid
+
+    def get_user(self, username):
```
Daniel-Sanche, line 81: Is there a reason why both services need access to all these functions? The design philosophy behind microservices seems to be to [avoid creating shared libraries](https://medium.com/standard-bank/microservices-dont-create-shared-libraries-2e803b033552) whenever possible. Is there a strong argument for coupling these services so tightly?

bharathkkb: @Daniel-Sanche yeah our initial plan with common lib was for noSQL+SQL(MySQL, PGSQL etc) but since now we are only targeting SQL I think we avoid creating on a common lib as SQLAlchemy will handle all SQL DBs.

Daniel-Sanche: Ok sounds good. Will you be updating this PR, or creating a new one?

bharathkkb: I will open a new PR

PR #103, opened 2020-04-09
Update load generator

DanSanche:
Resolves GitHub security alert about one of the dependencies


DanSanche:
That's the one. It said it couldn't create an auto-update for jinja because it would break other dependencies. I decided to just upgrade them all to make it simpler

DanSanche:
@murog just realized this update didn't even touch jinja... :facepalm:

Issue #52, opened 2020-03-25
Java tests: ledger-writer


murog:
## Description
Add unit tests to [ledger-writer](https://github.com/GoogleCloudPlatform/bank-of-anthos/tree/master/src/ledgerwriter) service

## Expected Behavior
- [x] Tests run in CI
- [ ] [LedgerWriterController](https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/156)
    - [x] versions()
    - [x]  readiness()
    - [x] refactors for testability
   -  [ ]  addTransaction
   -  [ ] validateTransaction
   -  [ ] submitTransaction

murog:
Re-scoped this issue to just ledger-writer

PR #108, opened 2020-04-13
Create unit tests for LedgerWriterController

meganzhao10:
Changelist:

- Create unit tests for methods versions() and readiness() in the LedgerWriterController class.
- Refactor LedgerWriterController constructor so that JWTVerifier and environment valuables can be injected during testing. Wrap JWTVerifier into a class and create a custom runtime exception for the class when it fails to generate the verifier.
- Create a public final variable READINESS_CODE for "ok" so that the unit tests can refer to the variable instead of the string.


meganzhao:
@googlebot I signed it!

Daniel-Sanche:
That licensecheck issue you're running into is checking to make sure all files start with the Google license header. You can use this tool to create them all at once: https://github.com/google/addlicense

src/ledgerwriter/src/test/java/anthos/samples/financedemo/ledgerwriter/LedgerWriterControllerTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Test
+    @DisplayName("Given the server is serving requests, return HTTP Status 200")
+    void readiness() {
+        // When
+        final String actualResult = ledgerWriterController.readiness();
+
+        // Then
+        assertNotNull(actualResult);
+        assertEquals(ledgerWriterController.READINESS_CODE, actualResult);
```
Daniel-Sanche, line 76: Shouldn't this test be checking for HttpStatus.OK?

Daniel-Sanche:
Great start! Can you add your tests to auto-run on each PR? You can add a block to `.github/workflows/smoke-tests.yml`, right after the Python Checkstyle block:

```    
- name: Java Unit Tests
      timeout-minutes: 20
      run: |
        mvn test
```

Daniel-Sanche:
Also, I suppose we should update `.github/workflows/README.md` to document the new tests, and possibly rename the smoke-tests.yml file now that it's not just running a smoke test

src/ledgerwriter/src/test/java/anthos/samples/financedemo/ledgerwriter/LedgerWriterControllerTest.java
meganzhao: The readiness() method in LedgerWriterController returns the string "ok" instead of HttpStatus.OK ("200 OK"). I wasn't sure about the rationale behind it, so I wrap "ok" into a constant variable instead of changing what the readiness() method returns. Should I change readiness() to either `return HttpStatus.OK` or `return new ResponseEntity<String>("ok", HttpStatus.OK)` then?

Daniel-Sanche: The `@ResponseStatus(HttpStatus.OK)` annotation on the method means that it will return status 200 by default. I think the "ok" string was just so it's easier to debug

I'm fine with keeping the "ok" string, or changing it to just `return HttpStatus.OK` with no message body if that's cleaner, but the status 200 is the important part that should be tested.

src/ledgerwriter/src/test/java/anthos/samples/financedemo/ledgerwriter/LedgerWriterControllerTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Test
+    void addTransaction() {
+        // todo: add tests to addTransaction
```
murog, line 81: Can you add an issue number to this, just to make sure we have the todo tracked somewhere 🙂 

You can create one or add reference to pr / file in comment on #52 

`TODO: [issue-52] add tests to addTransaction`

src/ledgerwriter/src/test/java/anthos/samples/financedemo/ledgerwriter/LedgerWriterControllerTest.java
```diff
@@ -0,0 +1,83 @@
...
+        environmentVariables = new EnvironmentVariables();
+        ledgerWriterController = new LedgerWriterController(verifier,
+                LOCAL_ROUTING_NUM, BALANCES_API_ADDR, VERSION);
+    }
+
+    @Test
+    @DisplayName("Given version number in the environment, " +
+            "return a ResponseEntity with the version number")
+    void version() {
+        // Given
```
murog, line 56: 👍 thanks for adding display names and comments to the tests

src/ledgerwriter/src/test/java/anthos/samples/financedemo/ledgerwriter/LedgerWriterControllerTest.java
meganzhao: Changed the return type to ResponseEntity and added test on status.

src/ledgerwriter/src/test/java/anthos/samples/financedemo/ledgerwriter/LedgerWriterControllerTest.java
meganzhao: Added the reference to the comment, thanks for pointing that out!

src/ledgerwriter/src/main/java/anthos/samples/financedemo/ledgerwriter/GenerateKeyException.java
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package anthos.samples.financedemo.ledgerwriter;
+
+public class GenerateKeyException extends RuntimeException {
+    public GenerateKeyException(String message, Exception e) {
+        super(message, e);
+    }
+}
```
Daniel-Sanche, line 23: nit: do you think we should put this class in the same file as the JWTVerifierGenerator? It's a small class that's only used in one place, and the internet seems to [say it's ok](https://www.quora.com/Can-we-keep-more-than-one-class-in-a-single-java-file). But I'm not too familiar with Java best practices

meganzhao: Since the exception is only being used by JWTVerifierGenerator, I think this makes sense. I added the exception as a static inner class of JWTVerifierGenerator.

PR #111, opened 2020-04-14
Add strict mode to script

ahmetb:
Otherwise a failed command won't cause script to terminate.

### DIFF
--- a/.github/workflows/install-dependencies.sh
+++ b/.github/workflows/install-dependencies.sh
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -euo pipefail
+
 # install git, kubectl, maven, jdk, pip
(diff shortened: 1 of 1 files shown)


ahmetb:
So the smoke tests fail on `grep Aggregated` which is actually indicating a test issue (i.e. string not found). Tests need to be better in this case.

Daniel-Sanche:
@ahmetb I re-ran the tests, so I can't see the previous error log. I opened https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/112, but this may become less relevant as we move to more robust UI tests (https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/53)

ahmetb:
Feel free to merge as I don't have commit rights. :)

ahmetb:
Yeah that makes sense. I think it simply was a timing issue.

Issue #51, opened 2020-03-25
Python tests: userservice


murog:
## Description
Add tests to [userservice](https://github.com/GoogleCloudPlatform/anthos-finance-demo/blob/master/src/userservice) 

## Expected Behavior
- [x] Tests run in CI
- [x] Test cases
   *  `/version` returns expected version with 200 OK
   * `/ready` returns 200 OK
   * can create user and return 201 and expected user
   * creating existing user returns 409 
   * sql error results in 500 and error message 
   *  invalid user with missing fields returns 400 
   *  invalid user with mismatched password values results in 400
   *  can log in with existing user
   * logging in with existing user and invalid password results in 404

murog:
removed frontend from needing tests since angular port is underway

murog:
re-scoped issue to just userservice to be consistent with java testing issues

murog:
fixed in #128

PR #128, opened 2020-04-15
Add tests for userservice and supporting refactor

bharathkkb:
Changelist
part of #51 

- Add tests for userservice
- Refactor db ops to a seperate module to increase testability
- Refactor Flask app to follow application factory pattern
-- enables users to run userservice locally using flask dev server `python3 userservice.py`
-- easier to create test instances


.github/workflows/smoke-tests.yml
```diff
@@ -45,8 +45,8 @@ jobs:
     - name: Python Checkstyle
       timeout-minutes: 20
       run: |
-        pylint --disable=F0401 --disable=W0511 --disable=E1101 --disable=R0801 --disable=R0903  ./src/*/*.py
-        # disabled warnings: import loading, todos, function members, duplicate code, public methods
+        pylint --disable=F0401 --disable=W0511 --disable=E1101 --disable=R0801 --disable=R0903 --disable=R0915 ./src/*/*.py
```
bharathkkb, line 48: pylint was throwing this warning since userservice.create_app > 50 lines. We can refactor userservice into a [blueprint](https://flask.palletsprojects.com/en/1.1.x/blueprints/) style approach but I feel that may make it unnecessarily complicated.

.github/workflows/smoke-tests.yml
```diff
@@ -45,8 +45,17 @@ jobs:
...
+        # disabled warnings: import loading, todos, function members, duplicate code, public methods, too many statements
+    - name: Python Unit Tests
+      timeout-minutes: 20
+      run: |
+        cd src/userservice
+        python3 -m venv env
+        source env/bin/activate
+        pip install -r tests/requirements-test.txt
+        python -m pytest -v -p no:warnings
+        deactivate
```
Daniel-Sanche, line 58: "set -x" can be useful to see more detailed logs as the test is run

.github/workflows/smoke-tests.yml
bharathkkb: Yes, more details from official docs here about [factories](https://flask.palletsprojects.com/en/1.1.x/patterns/appfactories/#factories-extensions) and usage with [gunicorn](https://flask.palletsprojects.com/en/1.1.x/deploying/wsgi-standalone/#gunicorn)

src/userservice/userservice.py
```diff
@@ -20,162 +20,165 @@
...
-    passhash = bcrypt.hashpw(password.encode('utf-8'), salt)
-
-    accountid = _generate_accountid()
-
-    # Add user to database
-    data = {'accountid': accountid,
+    app = Flask(__name__)
+
+    @app.route('/version', methods=['GET'])
+    def _version():
```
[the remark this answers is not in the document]
bharathkkb: route decorated funcs throws a[ lint error with pylint](https://github.com/pallets/flask/issues/2434) when they are inner. We can also disable this in pylint; I dont have a strong pref, but leaning towards removing underscore. Will do some research about best practice.

src/userservice/userservice.py
```diff
@@ -20,162 +20,165 @@
...
+            raise UserWarning('missing required field(s)')
+        if any(not bool(req[f] or req[f].strip()) for f in fields):
+            raise UserWarning('missing value for input field(s)')
+
+        # Check if passwords match
+        if not req['password'] == req['password-repeat']:
+            raise UserWarning('passwords do not match')
+
+    @app.route('/login', methods=['GET'])
+    def _get_token():
```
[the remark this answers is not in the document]
bharathkkb: Done

src/userservice/tests/test_userservice.py
```diff
@@ -0,0 +1,205 @@
...
+        self.flask_app.config['VERSION'] = version
+        # send get request to test client
+        response = self.test_app.get('/version')
+        # assert both versions are equal
+        self.assertEqual(response.data, version.encode())
+
+    def test_ready_endpoint(self):
+        """test if correct response is returned from readiness probe"""
+        response = self.test_app.get('/ready')
+        self.assertEqual(response.data, b'ok')
```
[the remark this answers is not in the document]
bharathkkb: Done

src/userservice/tests/test_userservice.py
```diff
@@ -0,0 +1,205 @@
...
+        self.assertEqual(response.status_code, 200)
+        # assert we get a json response with just token key
+        self.assertEqual(list(response.json.keys()), ['token'])
+        # assert token is example token
+        self.assertEqual(response.json['token'], EXAMPLE_TOKEN)
+
+    # mock check pw to return false
+    @patch('bcrypt.checkpw', return_value=False)
+    def test_login_invalid_password(self, _mock_checkpw):
+        """test logging in with existing user"""
```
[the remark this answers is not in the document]
bharathkkb: Fixed

src/userservice/tests/test_userservice.py
```diff
@@ -0,0 +1,205 @@
...
+        example_user_request = EXAMPLE_USER_REQUEST.copy()
+        self.mocked_db.return_value.get_user.return_value = example_user
+        # send request to test client
+        response = self.test_app.get('/login', query_string=example_user_request)
+        # assert 200 response
+        self.assertEqual(response.status_code, 200)
+        # assert we get a json response with just token key
+        self.assertEqual(list(response.json.keys()), ['token'])
+        # assert token is example token
+        self.assertEqual(response.json['token'], EXAMPLE_TOKEN)
```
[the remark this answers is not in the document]
bharathkkb: Added

src/userservice/tests/test_userservice.py
```diff
@@ -0,0 +1,205 @@
...
+        self.assertEqual(response.status_code, 500)
+        # assert we get correct error message
+        self.assertEqual(response.json['msg'], 'failed to create user')
+
+    def test_create_user_malformed(self):
+        """test creating a new user without required keys"""
+        # create example user request
+        example_user = EXAMPLE_USER_REQUEST.copy()
+        # remove a required field
+        example_user.pop('username')
```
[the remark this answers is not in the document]
bharathkkb: Added this - not sure if you meant use 1 object and keep removing keys from it or to try removing a different key each time. I implemented the latter since I thought it was a better test but can switch to former as well.

.github/workflows/smoke-tests.yml
bharathkkb: I am running it with -v which is the verbose flag, not sure if set -x is necessary

src/userservice/userservice.py
```diff
@@ -20,162 +20,165 @@
...
+
+        except LookupError as err:
+            return jsonify({'msg': str(err)}), 404
+        except PermissionError as err:
+            return jsonify({'msg': str(err)}), 401
+        except SQLAlchemyError as err:
+            app.logger.error(err)
+            return jsonify({'msg': 'failed to retrieve user information'}), 500
+
+    def __create_user_obj(user):
```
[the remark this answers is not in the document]
bharathkkb: I went with the first approach. Second approach will make db module have more scope like generating pwd hash. I have also modified the add user test to make it more comprehensive

src/userservice/tests/test_db.py
```diff
@@ -0,0 +1,91 @@
...
+        user['username'] = 'baz'
+        user['accountid'] = '3'
+        # add baz_user to db
+        self.db.add_user(user)
+        # get baz_user from db
+        db_user = self.db.get_user(user['username'])
+        # assert both user objects are equal
+        self.assertEqual(user, db_user)
+
+    def test_get_non_existent_user(self):
```
djmailhot, line 73: (nit) Consider using more descriptive names for your test methods
(small, for every unit test method)

A common best practice template I see is `functionname_situation_expectedresult`.  Obviously, long method names are expected.
For example: `test_get_nonexistent_user_returns_none`

Resources:
[Naming standards for unit tests](https://osherove.com/blog/2005/4/3/naming-standards-for-unit-tests.html)
[Python testing style guidelines](https://jrsmith3.github.io/python-testing-style-guidelines.html)
[Testing Your Code](https://docs.python-guide.org/writing/tests/)
etc, etc...

bharathkkb: great suggestion, helpful to see what failed without digging into test code.

src/userservice/userservice.py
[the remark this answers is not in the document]
Daniel-Sanche: Hmm that's tough. I think it's more readable without the underscore, but that seems like that could a useful pylint check in other contexts. Either way works for me

Do you think we should try `pylint-flask`?

src/userservice/tests/test_userservice.py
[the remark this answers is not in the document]
Daniel-Sanche: Yeah I was also thinking of the later, looks good!

.github/workflows/smoke-tests.yml
Daniel-Sanche: I was thinking of adding "set -x" as a bash command at the top, so we can see each command as they are run (like the License Check)

bharathkkb: done

src/userservice/userservice.py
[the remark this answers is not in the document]
bharathkkb: What about disabling inline? like so `def version():  # pylint: disable=unused-variable` with a comment explaining why we have done this.
I see that being followed in the [official google auth lib for python](https://github.com/GoogleCloudPlatform/google-auth-library-python-httplib2/blob/e7cd722281d1d897fa9ae6e3b6b78ae142778e6e/tests/compliance.py#L39) which uses flask and in [Tensorflow](https://github.com/google/TensorNetwork/blob/54574d53de888003ba8b184737ca39d615bd43ff/tensornetwork/backends/backend_test.py)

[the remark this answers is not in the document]
bharathkkb: @Daniel-Sanche a gentle ping for ^ my latest commits have this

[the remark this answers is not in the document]
Daniel-Sanche: Would we have to do that for every function? That seems distracting. It might be better to just disable the error globally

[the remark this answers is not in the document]
bharathkkb: @Daniel-Sanche yes, [right now its one per function,](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/5c71ae1b8ff84da4af91678197acc18bef314ebf/src/userservice/userservice.py#L56) we can also disable it just for this file if that is better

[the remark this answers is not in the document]
Daniel-Sanche: Yeah sure, that seems fine to me!

src/userservice/.dockerigone
```diff
@@ -0,0 +1 @@
+tests/*
```
murog, line 1: 👍

.github/workflows/smoke-tests.yml
```diff
@@ -45,12 +45,22 @@ jobs:
     - name: Python Checkstyle
       timeout-minutes: 20
       run: |
-        pylint --disable=F0401 --disable=W0511 --disable=E1101 --disable=R0801 --disable=R0903  ./src/*/*.py
-        # disabled warnings: import loading, todos, function members, duplicate code, public methods
+        pylint --disable=F0401 --disable=W0511 --disable=E1101 --disable=R0801 --disable=R0903 --disable=R0915 ./src/*/*.py
```
murog, line 48: not a change to this PR, but would it be possible to move this to a config in the future? would look cleaner and be easier to maintain.

.github/workflows/smoke-tests.yml
```diff
@@ -45,12 +45,22 @@ jobs:
...
+        # disabled warnings: import loading, todos, function members, duplicate code, public methods, too many statements
     - name: Java Unit Tests
       timeout-minutes: 20
       run: |
         mvn test
+    - name: Python Unit Tests
+      timeout-minutes: 20
+      run: |
+        set -x
+        cd src/userservice
```
murog, line 58: does each step restart from the project root, or does the directory change persist in the other steps?

src/userservice/tests/test_userservice.py
```diff
@@ -0,0 +1,242 @@
...
+                with patch('userservice.userservice.UserDb') as mock_db:
+                    self.mocked_db = mock_db
+                    # get create flask app
+                    self.flask_app = create_app()
+                    # set testing config
+                    self.flask_app.config['TESTING'] = True
+                    # create test client
+                    self.test_app = self.flask_app.test_client()
+
+    def test_version_endpoint_returns_200_status_code_correct_version(self):
```
murog, line 67: great descriptive names 👍

.github/workflows/smoke-tests.yml
bharathkkb: yes I think we can either make .pylintrc per python microservice or one for all

src/userservice/tests/test_userservice.py
```diff
@@ -0,0 +1,242 @@
...
+        example_user['username'] = ''
+        # send request to test client
+        response = self.test_app.post('/users', data=example_user)
+        # assert 400 response code
+        self.assertEqual(response.status_code, 400)
+        # assert we get correct error message
+        self.assertEqual(response.json['msg'], 'missing value for input field(s)')
+
+    def test_create_user_mismatch_password_400_status_code_error_message(self):
+        """test creating a new user with empty value for required key"""
```
murog, line 176: nit: think this should be `test creating a new user with mismatched password values`

.github/workflows/smoke-tests.yml
bharathkkb: Yes I believe so, unless overridden with `working-directory`
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsrun

src/userservice/tests/test_userservice.py
bharathkkb: Thanks to @djmailhot :D

src/userservice/tests/test_userservice.py
bharathkkb: Good catch, fixed

src/userservice/userservice.py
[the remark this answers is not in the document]
bharathkkb: Added it at the block level so it only affects that particular block.

.github/workflows/smoke-tests.yml
murog: created #136

PR #147, opened 2020-04-28
E2E: Added login and account creation tests

murog:
Part of Issue #53 

- [x] 1: User can log in with default account
   * be redirected to home
   * see correct username
   * see correct balance
   * see transaction history of at least 10
   * does not see messages for empty transaction or error
- [x] 2: User is blocked from signing in with bad credentials
  * fails with alert banner
  * cannot access home page
- [x] 3: User can create account
   * be redirected to home
   * see zero balance
   * see correct username
   * see empty transaction history message
  * sees no transaction list
- [x] 7: User is redirected from /home to /login when not authenticated
- [x] 8: User is redirected from /login and /signup to /home when already authenticated

Cypress.json changes
* moved defaultUser and messages to config

Frontend changes: 
* Updated account-user-name to be id
* added current-balance id
* added transaction-* id's
* added login-* id's
* added signup-* id's


.github/workflows/ui-tests/cypress/integration/login_spec.js
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+describe('Login Page', function() {
+  it('successfully loads', function() {
+    cy.visit('/login')
+  })
+})
+
+describe('Default Credentials on Form Submission', function() {
+  const username = 'testuser'
```
bharathkkb, line 24: @murog these should be pre filled based on `value="testuser"` right? Would it not be better to retrieve that value and use it so we know that the pre filling also works as intended.

.github/workflows/ui-tests/cypress/integration/login_spec.js
```diff
@@ -0,0 +1,79 @@
...
+  })
+
+  it('sees correct username', function() {
+    cy.get('#account-user-name').contains(name)
+  })
+
+  it('sees correct balance', function() {
+    cy.get("#current-balance").then(($span) => {
+      const balanceStr= $span.text()
+      const balance = parseFloat(balanceStr.replace(/[^\d.]/g, ''))
```
bharathkkb, line 43: @murog small nit: maybe a comment indicating what this regex does? I assume we get the dollar sign in the `balanceStr` and this extracts just the digits.

.github/workflows/ui-tests/cypress/integration/create_spec.js
```diff
@@ -0,0 +1,61 @@
...
+      cy.visit('/signup')
+      cy.get('.header-title').contains('Register a new account')
+    })
+})
+
+describe('User can create account', function() {
+    const uuid = () => Cypress._.random(0, 1e6)
+    const password = 'bells'
+    const firstName = 'Tom'
+    const lastName = 'Nook'
```
Daniel-Sanche, line 35: :raccoon:

.github/workflows/ui-tests/cypress/integration/create_spec.js
```diff
@@ -0,0 +1,61 @@
...
+    })
+
+    it('redirected to home', function() {
+        cy.url().should('include', '/home')
+    })
+
+    it('contain zero balance', function() {
+        cy.get('#current-balance').contains(expectedBalance)
+    })
+
```
Daniel-Sanche, line 57: Is it possible to check for an empty transactions list too?

.github/workflows/ui-tests/cypress/integration/login_spec.js
```diff
@@ -0,0 +1,79 @@
...
+      cy.wrap(balance).should('greaterThan', 0)
+     })
+  })
+
```
Daniel-Sanche, line 47: same as above: can we check for default transaction list?

.github/workflows/ui-tests/cypress/integration/login_spec.js
murog: good catch.  I think it's best to keep the two things decoupled, so a test for verifying the pre-filled username and password are what we expect and tests for login of our default credentials

.github/workflows/ui-tests/cypress/integration/login_spec.js
murog: since we can't predict what's in the transaction list with the current default user, added check that there are at least 10 rows

.github/workflows/ui-tests/cypress/integration/create_spec.js
```diff
@@ -0,0 +1,79 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+describe('User can navigate to create account screen', function() {
+    it('from login', function() {
+        cy.visit('/login')
+        cy.get('.btn-create-account').click()
```
ChandranshuRao14, line 20: nit: A cleaner approach might be to add an `id` to the signup button and use that in the selector rather than the `class`. This would prevent tests from breaking on style changes and keeps it consistent with the rest of the `cy.get` usage in these tests. [Cypress best practices](https://docs.cypress.io/guides/references/best-practices.html) recommends avoiding the use of class selectors as well. WYDT?

.github/workflows/ui-tests/cypress/integration/create_spec.js
```diff
@@ -0,0 +1,79 @@
...
+        cy.url().should('contain', 'signup')
+
+    })
+
+    it('from url', function() {
+      cy.visit('/signup')
+      cy.get('.header-title').contains('Register a new account')
```
ChandranshuRao14, line 27: Same as above

.github/workflows/ui-tests/cypress/support/commands.js
```diff
@@ -0,0 +1,45 @@
...
+    Cypress.log({
+        name: 'login',
+        message: `${username} | ${password}`,
+    })
+
+    cy.visit('login')
+
+    cy.get('input[name=username]').clear().type(username)
+    cy.get('input[name=password]').clear().type(password)
+    cy.get('form').submit()
```
ChandranshuRao14, line 27: We should probably use id here as well

.github/workflows/ui-tests/cypress/integration/create_spec.js
murog: great point, totally agree!

.github/workflows/ui-tests/cypress/integration/login_spec.js
```diff
@@ -0,0 +1,106 @@
...
+      const balanceStr= $span.text()
+      // regex: removes any characters that are not a digit [0-9] or a period [.]
+      const balance = parseFloat(balanceStr.replace(/[^\d.]/g, ''))
+      cy.wrap(balance).should('greaterThan', 0)
+    })
+  })
+
+  it('sees transaction history', function() {
+    cy.get('#transaction-table').children().should('have.class', 'table')
+    cy.get('.list').children().should('have.length.greaterThan', 10)
```
ChandranshuRao14, line 61: Oops just caught this, looks like this one was missed.

.github/workflows/ui-tests/cypress/integration/login_spec.js
Daniel-Sanche: sounds good!

Issue #151, opened 2020-04-28
Update CI to test current manifests


Daniel-Sanche:
The current CI system tests the latest manifests against the latest code. We also want to check against the latest published images to ensure `kubectl apply -f` continues to work

https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/132

PR #153, opened 2020-04-28
Add test matrix

Daniel-Sanche:
We had a blind spot in our tests, where the manifests on master might not be compatible with the latest published images. This should fix that scenario, so we know to push minor version releases when the manifests change significantly enough

https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/151

- Add test matrix to test release manifests and latest code concurrently
- added skaffold profile to make adding more tests to the matrix easier in the future
- moved code tests into separate job so for better parallelization
- changed name of smoke-tests.yaml, since it has expanded in scope
- updated README for GitHub workflows
- modified release script to allow creating releases from other branches


Daniel-Sanche:
Ignore the yellow test status - I updated the test names, so I'll have to update the required tests in the repo settings after this is approved

.github/workflows/ci.yaml
```diff
@@ -60,28 +60,44 @@ jobs:
...
+      fail-fast: true
+    steps:
+    - uses: actions/checkout@v2
     - name: Setup Cluster
       run: |
         kind delete cluster 2>/dev/null
         kind create cluster
         kubectl get nodes
     - name: Deploy From Source
+      if: ${{ matrix.profile != 'latest-release' }}
```
djmailhot, line 77: `matrix.profile != 'latest-release'`  ->>  `matrix.profile == 'local-code'`

.github/workflows/ci.yaml
```diff
@@ -60,28 +60,44 @@ jobs:
         pip install -r tests/requirements-test.txt
         python -m pytest -v -p no:warnings
         deactivate
+  deployment-tests:
+    runs-on: self-hosted
+    strategy:
+      matrix:
```
murog, line 66: reading up on actions matrix, very cool! could be helpful way of thinking about future environments and profiles

.github/workflows/ci.yaml
```diff
@@ -60,28 +60,44 @@ jobs:
...
-        kubectl wait --for=condition=ready --timeout=500s pod/accounts-db-0
-        kubectl wait --for=condition=ready --timeout=500s pod/ledger-db-0
         kubectl wait --for=condition=available --timeout=500s deployment/frontend
         kubectl wait --for=condition=available --timeout=500s deployment/contacts
         kubectl wait --for=condition=available --timeout=500s deployment/userservice
         kubectl wait --for=condition=available --timeout=500s deployment/ledgerwriter
         kubectl wait --for=condition=available --timeout=500s deployment/transactionhistory
         kubectl wait --for=condition=available --timeout=500s deployment/balancereader
         kubectl wait --for=condition=available --timeout=500s deployment/loadgenerator
+        kubectl wait --for=condition=ready --timeout=500s pod/accounts-db-0
```
murog, line 99: qq: were these moved for functional or formatting reasons?

.github/workflows/ci.yaml
```diff
@@ -99,19 +115,18 @@ jobs:
...
         }
         port_forward &
     - name: Ping frontend
       run: |
-          curl --silent --output /dev/null --show-error --fail localhost:3000 
+          curl --silent --output /dev/null --show-error --fail localhost:3000
     - name: UI tests
       run: |
-        docker run -t -v $PWD/.github/workflows/ui-tests:/e2e -w /e2e --network=host cypress/included:4.3.0
+          docker run -t -v $PWD/.github/workflows/ui-tests:/e2e -w /e2e --network=host cypress/included:4.3.0
```
murog, line 132: thanks for fixing all these whitespace nits!

.github/workflows/ci.yaml
```diff
@@ -24,7 +24,7 @@ on:
     branches:
       - master
 jobs:
-  run-tests:
+  code-tests:
```
murog, line 27: great idea to split up the jobs! if the deployment jobs has a flaky fail in the future, only the deployment job needs to be rerun right?

bharathkkb: I encountered this a couple of months before and I think its still a FR on Github side
Source: https://github.community/t5/GitHub-Actions/Ability-to-rerun-just-a-single-job-in-a-workflow/td-p/41629

.github/workflows/ci.yaml
Daniel-Sanche: I was trying to leave it open to add more skaffold profiles for testing that would all use this step (maybe for deploying and testing the monolith?)

.github/workflows/ci.yaml
Daniel-Sanche: Surprisingly, functional. "Deploy Latest Release" fails if this is on top

The db manifests create a statefulset, but you can't wait on that directly, you have to wait on the underlying Pod. But the Pod takes a couple seconds to appear, and if you wait on it before it exists, it will return an error

.github/workflows/ci.yaml
Daniel-Sanche: Yeah, I think you have to re-run it all. The code tests are super quick though, so re-running isn't a huge deal

murog: 😭

.github/workflows/ci.yaml
murog: well that sounds like it was fun to debug 😄 thanks for explaining

.github/workflows/ci.yaml
```diff
@@ -60,28 +60,44 @@ jobs:
         pip install -r tests/requirements-test.txt
         python -m pytest -v -p no:warnings
         deactivate
+  deployment-tests:
```
bharathkkb, line 63: Using [Job ordering](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) I think we should enforce the job dependencies like if `code-tests `fail, we should not need to proceed to `deployment-tests.` 

Right now from what I understand `code-tests` and 2 `deployment-tests` jobs all start in parallel (if there are enough runners).  In the event that `code-tests` fail, we still need to wait till one of the `deployment-tests` to fail before the whole workflow is complete and considered failed. 

With job ordering we can immediately fail if `code-tests` fail. Maybe this isn't needed, but interested to know what you think.

.github/workflows/ci.yaml
bharathkkb: +1

.github/workflows/ci.yaml
Daniel-Sanche: I definitely tried a couple hacky workarounds before realizing I could just move this line to the bottom!

.github/workflows/ci.yaml
Daniel-Sanche: Hmm I guess you're right. I was thinking `fail-fast` should cover us, but I guess that only covers the matrix. I'll add a `needs` statement to do the code checks first. Good suggestion!

Daniel-Sanche:
I had to release v.0.1.2 for this, since the published manifests weren't passing the new UI tests. It was a bit clunky, we may want to document this better for contributors

PR #159, opened 2020-05-06
Hardcode account numbers in default data scripts instead of using environment variables.

### LINKED ISSUE #131: Don't use environment variables to define default database data

djmailhot:
Resolves #131 

Hardcode account numbers in default data scripts instead of using environment variables.

 1. Create 3 user accounts:  'alice', 'bob', and 'eve'. (password for every account is 'password')
 2. Set everyone as contacts of each other.
 3. Have all 3 accounts send each other payments, as well as to random anonymous accounts.

See screenshot for example transaction history for 'eve':
transactions


kubernetes-manifests/accounts-db.yaml
```diff
@@ -33,8 +33,9 @@ spec:
         tier: db
     spec:
       containers:
-      - name: accounts-db
+      - name: postgres
         image: gcr.io/bank-of-anthos/accounts-db:v0.1.1
+        imagePullPolicy: "IfNotPresent"
```
Daniel-Sanche, line 38: Was there a reason for adding this?

kubernetes-manifests/config.yaml
```diff
@@ -39,13 +39,4 @@ data:
...
-  DEFAULT_USER_USERNAME: "testuser"
-  DEFAULT_USER_NAME: "Eve"
-  DEFAULT_DEPOSIT_ACCOUNT: "9099791699"
-  DEFAULT_DEPOSIT_ROUTING: "808889588"
-  DEFAULT_DEPOSIT_LABEL: "External Bank"
-  DEFAULT_CONTACT_ACCOUNT_A: "1044226144"
-  DEFAULT_CONTACT_NAME_A: "Alice"
-  DEFAULT_CONTACT_ACCOUNT_B: "1055757655"
-  DEFAULT_CONTACT_NAME_B: "Bob"
+  DEFAULT_LOGIN_USERNAME: "eve"
```
Daniel-Sanche, line 42: So no more testuser? Isn't Eve usually the evil one?

kubernetes-manifests/config.yaml
```diff
@@ -39,13 +39,4 @@ data:
   USE_DEFAULT_DATA: "True"
   # All default user accounts are hardcoded to use the login password 'password'
   DEFAULT_LOGIN_PASSWORD: "password"
```
Daniel-Sanche, line 41: Are these `DEFAULT_LOGIN` variables just for the frontend autofill? If so, we should probably rename them to make it clearer, and/or move them to just the frontend service instead of the shared configmap

kubernetes-manifests/accounts-db.yaml
djmailhot: It's in ledger-db.yaml.  Wanted to make things consistent.

kubernetes-manifests/config.yaml
djmailhot: That's fair.  Hmm, maybe testuser can be friends with alice, bob, and eve?

Daniel-Sanche:
Looks like you're going to have issues with the published images along with this PR. You'll need to [cut a minor version release](https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/161) for the tests to pass

Let me know what you think of that process. At this point, I'm kind of leaning towards just using `:latest` as the tag instead of tagging to outdated images. I don't want these minor releases to have to be a common thing

kubernetes-manifests/config.yaml
Daniel-Sanche: That sounds good to me. 3 contacts might look better in the history list too, more variety. If it's a lot of work, this isn't a big issue for me though

kubernetes-manifests/accounts-db.yaml
djmailhot: Apparently [IfNotPresent is the default when imagePullPolicy is not set](https://kubernetes.io/docs/concepts/containers/images/#updating-images)

Removing.

kubernetes-manifests/config.yaml
djmailhot: I got a little carried away and renamed everything "Default Data" to "Demo Data".  Let me know if you think that makes the purpose clearer.

kubernetes-manifests/config.yaml
djmailhot: Done and done.  Easy peasy with these new script files.

murog:
FYI in pr #183, i use the default user's [external accounts and recipients](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/e1345a56f711adc0ad48680d0aff48b38eaa938c/.github/workflows/ui-tests/cypress.json)

Daniel-Sanche:
@djmailhot is this PR still relevant?

djmailhot:
> @djmailhot is this PR still relevant?

Yep still hoping to get this in.  Hard coding the numbers will play much better with the Java monolith.

Re #183 , I will remember to update the test attributes.

src/ledger-db/initdb/1_create_transactions.sh
```diff
@@ -14,53 +14,98 @@
...
+    for i in $(seq 1 $PAY_PERIODS); do
+        # create deposit transaction for each user
+        for account in ${USER_ACCOUNTS[@]}; do
+            add_transaction "$EXTERNAL_ACCOUNT" "$account" "$EXTERNAL_ROUTING" "$LOCAL_ROUTING_NUM" $DEPOSIT_AMOUNT $START_TIMESTAMP
+        done
 
-# create a UNIX timestamp in seconds since the Epoch
-START_TIMESTAMP=$(( $(date +%s) - $(( $(($PAY_PERIODS+1)) * $SECONDS_IN_PAY_PERIOD  ))  ))
+        # create payments between users
+        TRANSACTIONS_PER_PERIOD=$(shuf -i 10-20 -n1)
```
Daniel-Sanche, line 65: If I'm understanding this right, if you get extremely unlucky and select the same `SENDER_ACCOUNT` for 20 payments each of $250, you'll end up spending $5000. This is more than the $2500 deposited in a pay period, so you could potentially create a negative balance.

Maybe you should adjust the numbers or logic so that this isn't possible?

djmailhot: Good point.  Adjusted so that an account will never be overdrawn.

PR #170, opened 2020-05-15
Don't hard code zone

justingrayston:
This should be able to be set to demo in appropriate region.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,3 +43,7 @@ check-env:
 ifndef PROJECT_ID
 	$(error PROJECT_ID is undefined)
 endif
+
+ifndef ZONE
+	$(error ZONE is undefined)
+endif
--- a/README.md
+++ b/README.md
@@ -48,8 +48,9 @@ cd bank-of-anthos
 ### 3 - Create a Kubernetes cluster
 
 ```
+ZONE=<your-zone>
 gcloud beta container clusters create bank-of-anthos \
-    --project=${PROJECT_ID} --zone=us-central1-b \
+    --project=${PROJECT_ID} --zone=${ZONE} \
     --machine-type=n1-standard-2 --num-nodes=4
 ```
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -49,7 +49,7 @@ cd bank-of-anthos
 
 ```
```
djmailhot, line 50: Add a line in this step to define the zone, similar to `PROJECT_ID=<your-project-id>` above

```
ZONE=<your-zone>
```

justingrayston: Pushed. Please take a look when ever you get a mo. Have a great weekend!

djmailhot: Actually, I think this might work better here under step 3 **Create a Kubernetes cluster**.  That is:

```
ZONE=<your-zone>
gcloud beta container clusters create bank-of-anthos \
    --project=${PROJECT_ID} --zone=us-central1-b \
    --project=${PROJECT_ID} --zone=${ZONE} \
    --machine-type=n1-standard-2 --num-nodes=4
```

Makefile
```diff
@@ -43,3 +43,7 @@ check-env:
 ifndef PROJECT_ID
 	$(error PROJECT_ID is undefined)
 endif
+
+ifndef ZONE
+	$(error ZONE is undefined)
```
Daniel-Sanche, line 48: I'd actually vote against this addition FWIW. I think it's better to provide a default that can be overwritten instead of requiring input for variables like this

But that's just me, feel free to merge either way

justingrayston: @Daniel-Sanche on line 17 a default is provided for ZONE and 18 for CLUSTER. I can only imagine these checks are put here just in case a user removes them? 🤷‍♂️

README.md
justingrayston: Yes, I agree. That flows better...pushed.

justingrayston:
@djmailhot I also noted the number of the steps was also incorrect, not sure how or from what branch, but I corrected.

Should be good now if you are happy.

Makefile
Daniel-Sanche: previously the only check was only for PROJECT_ID, which is the one required variable we can't set a default for.

Yeah I guess this new check doesn't really change anything though unless the user removes the default, so I'm fine either way

justingrayston:
ooops. What did I break!

PR #172, opened 2020-05-19
Limit username charset + length

### LINKED ISSUE #129: Add limitations to username
- limit number of characters
- limit characterset

validate on frontend and backend

ChandranshuRao14:
Fixes #129 

* Uses regex `([a-zA-Z0-9_]){2,15}` to match usernames with 2-15 chars and only alphanumeric or underscore characters
* Added in the HTML `pattern` attribute on the `username` inputs for the signup and login pages
* Validated in `userservice` backend


Daniel-Sanche:
mind adding a test?

ChandranshuRao14:
> mind adding a test?

Sure thing! Would a test in userservice suffice or should we add a UI test as well?

bharathkkb:
@ChandranshuRao14 I think it makes sense to add one test for userservice too. We had some regex related testing for contact service and you can leverage the same approach.
Example constants: https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/e715e4e445217d016a672a561b1530754abf1d4f/src/contacts/tests/constants.py#L75
Example test: https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/e715e4e445217d016a672a561b1530754abf1d4f/src/contacts/tests/test_contacts.py#L231

ChandranshuRao14:
Sorry for the delay on this. Will add a test case to `create_spec.js` as well as `userservice.py` and update this PR by the end of this week!

Daniel-Sanche:
This should be ready for a fresh review:
- I added the missing unit tests, and finished Anshu's work on the UI tests
- I had to change the Python regex expression slightly to fix a bug that was caught by the tests
- The Python tests were taking ~16 minutes for some reason, so I adjusted the CI script to save the virtualenv in the $HOME directory, so it doesn't have to download packages fresh on each run. I'm not sure why that problem just came up now

.github/workflows/ui-tests/cypress/integration/create_spec.js
```diff
@@ -28,6 +28,53 @@ describe('User can navigate to create account screen', function() {
     })
 })
 
+describe('Signup is unsuccessful with long username', function() {
+    const invalidFeedback = Cypress.env('messages').invalidFeedback
```
murog, line 32: since this is used in multiple describe blocks, it makes sense to move to import at beginning of file
example: https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/5ae1f311252d0bf85a1cd65656fb84485cbb8952/.github/workflows/ui-tests/cypress/integration/transfer_spec.js#L24

.github/workflows/ui-tests/cypress/integration/create_spec.js
```diff
@@ -28,6 +28,53 @@ describe('User can navigate to create account screen', function() {
     })
 })
 
+describe('Signup is unsuccessful with long username', function() {
```
murog, line 31: could be helpful to update test name or add comment for length

.github/workflows/ui-tests/cypress/integration/create_spec.js
```diff
@@ -28,6 +28,53 @@ describe('User can navigate to create account screen', function() {
...
+    const invalidFeedback = Cypress.env('messages').invalidFeedback
+
+    beforeEach(function() {
+        const user = {
+            username: '0123456789abcdefghijklmnopqrstuvwxyz',
```
murog, line 36: making the username random would keep it consistent with other tests, you could do a random string of length or a random number of at least 15 digits. 
example:
```javascript
username: randomString(15, 100)
or
username: invalid_user_${randomNumber(2,100)}
```
what do you think?

src/frontend/templates/signup.html
```diff
@@ -64,9 +64,9 @@ <h4 class="mb-4">Account Information</h4>
                             <div class="input-group-prepend mr-2">
                               <span class="input-group-text"><span class="material-icons">account_circle</span></span>
                             </div>
-                            <input class="form-control" type="text" id="signup-username" name="username" autocomplete="off" required>
+                            <input class="form-control" type="text" id="signup-username" name="username" autocomplete="off" pattern="([a-zA-Z0-9_]){2,15}" required>
                             <div class="invalid-feedback">
-                                Please enter a username.
+                                Please enter a valid username. Username must be 2 to 15 characters in length and contain only alphanumeric or underscore characters.
```
murog, line 69: 👍

.github/workflows/ui-tests/cypress/integration/create_spec.js
```diff
@@ -28,6 +28,53 @@ describe('User can navigate to create account screen', function() {
...
+        cy.get('.invalid-feedback').contains(invalidFeedback.username)
+    })
+})
+
+describe('Signup is unsuccessful with non-alphanumeric username', function() {
+    const invalidFeedback = Cypress.env('messages').invalidFeedback
+
+    beforeEach(function() {
+        const user = {
+            username: 'Tom@Nook',
```
murog, line 59: same as above, making this random would make it consistent

src/userservice/tests/constants.py
```diff
@@ -27,6 +28,9 @@ def generate_rsa_key():
     public_key = key.publickey().export_key()
     return private_key, public_key
 
+def get_random_string(length):
+    """Generate random string of given length"""
+    return ''.join(random.choice(string.ascii_lowercase) for i in range(length))
```
murog, line 33: 👍

src/userservice/tests/constants.py
```diff
@@ -70,3 +74,20 @@ def generate_rsa_key():
...
+# Usernames must be >1 and <=15 chars, alphanumeric and underscores
+INVALID_USERNAMES = [
+    None, # null
+    "", # empty string
+    " ", # only space
+    "b", # single character
+    " user", # starting with space
+    "*$&%($", # non alphanumeric characters
+    "user*new", # alphanumeric with non alphanumeric characters
+    "🏦💸", # emojis
```
murog, line 87: 💸

Daniel-Sanche:
@googlebot I consent

.github/workflows/ui-tests/cypress/integration/create_spec.js
Daniel-Sanche: That's a good point, but I think a test with lengths could be dangerous (what if it only fails with 16 characters, so 1 out of 86 tests fail and we assume it's just flakiness?)

I could do a random string of 16 digits, since that's the first failure case. Or is it possible to loop with multiple values for the length? is randomString defined somewhere, or do I need to implement it myself?

.github/workflows/ui-tests/cypress/integration/create_spec.js
Daniel-Sanche: What kind of randomness are you thinking here? I'd be cautious about trying random non-alphanumeric characters, because we want the tests to be consistent.

This would be another good use of a loop if we wanted to test a bunch of bad strings. But that should be covered in the unit tests at least

.github/workflows/ui-tests/cypress/integration/create_spec.js
murog: that's fair!  in that case, you can keep as is or just attach random number that would result in a 16 char username (keeping as is seems simplest 😄 )

random num example:
https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/f1039dc57df89aa74e98ec770e01819ac117a08f/.github/workflows/ui-tests/cypress/integration/create_spec.js#L31

```javascript
describe('Signup is unsuccessful with long username', function() {
    const invalidFeedback = Cypress.env('messages').invalidFeedback
    const uuid = () => Cypress._.random(1e7, 1e8)
    const longUsername = `invalid_${uuid}`
    beforeEach(function() {
        const user = {
            username: longUsername
            firstName: 'Tom',
            lastName: 'Nook',
            password: 'bells'
        }
        cy.createAccount(user)
    })

```

.github/workflows/ui-tests/cypress/integration/create_spec.js
murog: no worries,  since it's covered in unit test already leaving as is seems reasonable

Issue #182, opened 2020-05-20
UI Test: cannot contain more than 2 decimal digits


murog:
## Description
Test case: Transfer / Deposit cannot contain more than 2 decimal digits
Manual testing shows that the UI is exhibiting correct behavior.  However, Cypress UI tests fail and round to `5.01`

## Expected behavior
Test reflects:
![Screen Shot 2020-05-20 at 1 47 07 PM](https://user-images.githubusercontent.com/12957039/82495761-749b4880-9aa0-11ea-8641-a34c064b713d.png)
```
    it('cannot contain more than 2 decimal digits', function () {
        const invalidPayment = `5\.02\.35\.459`
        cy.transfer(recipient, invalidPayment)
        cy.get('.invalid-feedback').should('be.visible')
        cy.get('.invalid-feedback').contains(invalidFeedback.payment)
```

### Current behavior
![Transfer is unsuccessful with invalid data -- cannot contain more than 2 decimal digits (failed)](https://user-images.githubusercontent.com/12957039/82495628-3dc53280-9aa0-11ea-95e3-13658e56359e.png)

NimJay:
Update: This will be fixed in #548.
We should be able to resolve sometime in January (with @Shabirmean's help, when he's back in the office). :)
(Commenting to cool down this ticket from our dashboard.)

PR #183, opened 2020-05-20
E2E: Add Deposit and Transfer Tests

murog:
Fixes issue #53 
- [x] 4: User can deposit funds
    * See balance update
    * See transaction in history
    * See new contact show up
- [x] 5: User can transfer funds
    * See balance update
    * See transaction in history
    * See new contact show up
- [x] 6: User is blocked from sending invalid data
    * Amount > balance
    * Amount <= 0
    * TODO: Amount with more than 2 decimal digits issue --> created #182 
    * Badly formatted routing or account numbers

Cypress.json changes
* Added messages and default user accounts and recipients

Frontend changes
* added id's to `send payment` and `deposit funds` spans

🚨 Note! The latest release build is failing since the frontend is updated in this pr


Daniel-Sanche:
Your tests should pass once this is merged: https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/194

.github/workflows/ui-tests/cypress/integration/deposit_spec.js
```diff
@@ -0,0 +1,192 @@
...
+        cy.get('#depositFunds').should('be.visible')
+    })
+
+    it('sees expected external accounts', function () {
+        cy.get('#depositSpan').click()
+        cy.get('#depositFunds').should('be.visible')
+        cy.get('#accounts').children().first().as('firstOption')
+        cy.get('@firstOption').contains('External Bank')
+        cy.get('@firstOption').contains('9099791699')
+        cy.get('@firstOption').contains('808889588')
```
Daniel-Sanche, line 77: I'm not sure if we need to test exact values here. I'd say it's more important that there's a valid number, instead of asserting exactly what the number is. What do you think?

.github/workflows/ui-tests/cypress/integration/deposit_spec.js
```diff
@@ -0,0 +1,192 @@
...
+        cy.get('.alert').contains(depositMsgs.success)
+    })
+
+    it('can see balance update after deposit', function () {
+        const depositAmount = validPayment()
+        let expectedBalance
+        cy.get("#current-balance").then(($span) => {
+            const currentBalanceSpan = $span.text()
+            // regex: removes any characters that are not a digit [0-9] or a period [.]
+            const currentBalance = parseFloat(currentBalanceSpan.replace(/[^\d.]/g, ''))
```
Daniel-Sanche, line 93: maybe we should only remove , and $ if those are all we expect to see?

.github/workflows/ui-tests/cypress/integration/deposit_spec.js
```diff
@@ -0,0 +1,192 @@
...
+        cy.reload()
+        cy.get('#depositSpan').click()
+        cy.get('#accounts').contains(newExternalAccount.contactLabel)
+        cy.get('#accounts').contains(newExternalAccount.accountNum)
+        cy.get('#accounts').contains(newExternalAccount.routingNum)
+    })
+
+})
+
+describe('Deposit is unsuccessful with invalid data', function () {
```
Daniel-Sanche, line 145: maybe we should add a test deposit using the local routing number? (I'm actually not sure how that's handled internally. It seems to trigger the frontend bug we already know about)

.github/workflows/ui-tests/cypress/integration/deposit_spec.js
```diff
@@ -0,0 +1,192 @@
...
+
+        const paymentAmount = validPayment()
+
+        cy.depositToNewAccount(invalidExternalAccount, paymentAmount)
+        cy.get('.invalid-feedback').should('be.visible')
+        cy.get('.invalid-feedback').contains(invalidFeedback.accountNum)
+    })
+
+    it('cannot reference invalid routing number', function () {
+        const invalidExternalAccount = {
```
Daniel-Sanche, line 180: it might be interesting to loop through a bunch of invalid values

.github/workflows/ui-tests/cypress/integration/transfer_spec.js
```diff
@@ -0,0 +1,213 @@
...
+        cy.get('.alert').contains(transferMsgs.success)
+        cy.get('#paymentSpan').click()
+        cy.get('#payment-accounts').contains(newRecipient.contactLabel)
+        cy.get('#payment-accounts').contains(newRecipient.accountNum)
+
+    })
+
+})
+
+describe('Transfer is unsuccessful with invalid data', function () {
```
Daniel-Sanche, line 152: maybe add payment to self?

.github/workflows/ui-tests/cypress/integration/deposit_spec.js
murog: hmm I added it as a sanity check that the test data was correct.  how about changing it to reference the test data specifically, so example: `contains(externalAccount.accountNum)`

Daniel-Sanche: Yeah that would probably be best! But if that's too much work I don't think it's super important

Daniel-Sanche: That might be better so we only need to update the values in one place if things change. But I don't think this is too important

.github/workflows/ui-tests/cypress/integration/deposit_spec.js
murog: when cypress retrieves the span there are newline characters that had to be removed as well

.github/workflows/ui-tests/cypress/integration/deposit_spec.js
murog: what is the default user's local routing number?

murog: nvm found it https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/kubernetes-manifests/config.yaml

.github/workflows/ui-tests/cypress/integration/deposit_spec.js
murog: I think it would be best to hold off on that until screenshots are enabled because of how cypress errors are reported in log

Daniel-Sanche:
What do you think of removing the smoke test in this PR? It's been pretty flaky lately

Issue #173, opened 2020-05-20
Create monolith version of the 3 Java services


askmeegs:

djmailhot:
Implemented as the `ledgermonolith` service under `/src/ledgermonolith`.

PR #195, opened 2020-05-21
Add Java monolith ledgermonolith with deployment and startup scripts

djmailhot:
Add the ledgermonolith service to be deployed on a GCE instance.

Usage:  `make monolith`


djmailhot:
Scripting the creation / updating / deletion of GCP resources is _really_ annoying via gcloud.  This may be involved enough to justify using tooling like Terraform...

djmailhot:
NOTE:  took out the `/health` REST endpoint for now.

Both TransactionHistoryController.java and BalanceReaderController.java have the `/health` endpoint.  I consolidated other duplicate endpoints (e.g. `/version`) into LedgerMonolithController.java.  But the `/health` endpoint is more difficult because it checks the status of a background thread present in each TransactionHistoryController.java and BalanceReaderController.java.  Meaning that consolidating it into LedgerMonolithController.java won't allow for checking the background threads in the other controller classes.

Currently the `/health` endpoint is not used by our infrastructure, so this seems workable for now.  But I'm open to ideas on how to make this work.

djmailhot:
Resolves #173

src/ledgermonolith/README.md
```diff
@@ -0,0 +1,91 @@
...
+  - `HISTORY_LIMIT`
+    - the number of past transactions to store for each user
+    - optional. Defaults to 100
+  - `EXTRA_LATENCY_MILLIS`
+    - add fake extra latency in milliseconds to transaction history requests
+    - optional. Defaults to 0
+
+### Scripts
+
+- `scripts/push-artifacts.sh`: pushes build artifacts to Google Cloud Storage
```
askmeegs, line 68: I think it's a good idea to push "canonical"  monolith artifacts to Cloud Storage in the bank-of-anthos project, in the same way we push images to GCR in that project. Then provide instructions for users to use the pre-built artifacts.

src/ledgermonolith/README.md
```diff
@@ -0,0 +1,91 @@
...
+### Make
+
+```
+# In the root directory of the project repo
+PROJECT_ID=<your-project-id>
+ZONE=<your-gcp-zone>
+make monolith
+```
+
+### Bash
```
askmeegs, line 83: We should also provide deployment instructions for the rest of bank of anthos + monolith. 
- Manifests for all services except 3 java backends 
- Update frontend environment vars to point to ledgermonolith instead of 3 java backends. 

maybe a separate k8s manifests directory in this dir.

src/ledgermonolith/init/startup-script.sh
```diff
@@ -0,0 +1,113 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Startup script to start the ledgermonolith service from a JAR.
+#
+# Expects build artifacts to be available on Google Cloud Storage at
+# gs://bank-of-anthos/monolith.
```
askmeegs, line 20: This is problematic because if we're providing instructions for any user to `make monolith` and push to their project, hardcoding the storage bucket to the release bucket doesn't make sense.

We should create a dynamic storage bucket name based on the PROJECT_ID.

askmeegs:
> Scripting the creation / updating / deletion of GCP resources is _really_ annoying via gcloud. This may be involved enough to justify using tooling like Terraform...

@djmailhot please file an issue for this.

src/ledgermonolith/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+```
+# In the root directory of the project repo
+PROJECT_ID=<your-project-id>
+ZONE=<your-gcp-zone>
+./src/ledgermonolith/scripts/push-artifacts.sh
+./src/ledgermonolith/scripts/deploy-monolith.sh
+```
```
askmeegs, line 91: Please add instructions for how a user can check if the monolith is running - eg. ssh into the VM, what systemd / other commands to run to see logs + monolith status.

src/ledgermonolith/src/main/resources/banner.txt
```diff
@@ -0,0 +1,7 @@
+   _                 _     __ _                           __  __                             _       _      _      _      
+  | |      ___    __| |   / _` |   ___      _ _     o O O|  \/  |   ___    _ _      ___     | |     (_)    | |_   | |_
```
askmeegs, line 2: Please remove this. The java headers don't format properly in Cloud Logging.

src/ledgermonolith/scripts/push-artifacts.sh
```diff
@@ -0,0 +1,53 @@
...
+else
+  echo "PROJECT_ID: ${PROJECT_ID}"
+  echo "ZONE: ${ZONE}"
+fi
+
+CWD="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+
+# If the GCS bucket doesn't exist, then create it
+gsutil ls -p $PROJECT_ID gs://bank-of-anthos &> /dev/null
```
askmeegs, line 39: make GCS bucket name dynamic based on ${PROJECT_ID}, do not hardcode to bank-of-anthos bucket

src/ledgermonolith/README.md
djmailhot: Done.

src/ledgermonolith/README.md
djmailhot: Alrighty, sounds good.  Working on it.

djmailhot: Alrighty, finished up alternative deployment option for the rest of the app.
Instructions found at:  https://github.com/GoogleCloudPlatform/bank-of-anthos/tree/monolith#variant-ledger-monolith-service

The current approach is a little hacky.  I'm currently looking into the option of using Kustomize instead.

Issue #199, opened 2020-05-29
Deposit resulting in "sender not authenticated"


murog:
### Steps to Reproduce
1. Followed steps to create new BoA deployment
1. Created new account
1. Make deposit to new external account without account label

🚨 Flaky, only happened the first three times
### Expected Behavior
![image](https://user-images.githubusercontent.com/12957039/83214722-1e6e7b00-a11a-11ea-9f3f-70fdcb1f2459.png)

### Actual Behavior
![image](https://user-images.githubusercontent.com/12957039/83214681-07c82400-a11a-11ea-8f26-005f38df06e8.png)

murog:
If not reproducible could be helpful to have short documentation on revolving common error messages?

askmeegs:
TODO - check service logs in userservice

murog:
[message source: lederwriter](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/e67011bdba45144a99fd70dfbcb48e12c4d1f137/src/ledgerwriter/src/main/java/anthos/samples/financedemo/ledgerwriter/TransactionValidator.java#L81)

murog:
haven't been able to reproduce or find information in logs, closing for now

Daniel-Sanche:
I'm thinking this may have something to do with the token expiring. The tokens are set to [expire after an hour](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/4bdc25daf21a5f8b7a945380b1be9a60626e7de9/kubernetes-manifests/userservice.yaml#L45), and then the user should be sent back to the login page, but maybe there's a bug in there somewhere. Might be worth exploring at some point to see if we can reproduce this

murog:
I was able to reproduce again
1. Followed steps to create new BoA deployment
1. Login with default user
1. Make deposit without contact label

Daniel-Sanche:
Ok, I found the issue! This is a weird one that hits a weird edge case:

- 123456789 is the bank of Anthos routing number. By using that as the "external deposit" routing number, our backend is treating it as an internal transaction
- your account isn't 1234567890, but now you're trying to initiate an internal transaction on their behalf, so the backend rejects it

The LedgerWriter is actually kind of behaving properly in this case. I'll add an extra check in the frontend to give a more appropriate error message though

PR #224, opened 2020-05-30
Reject deposits with local routing number

Daniel-Sanche:
fixes https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/199

TL;DR: give a better error message when users try to initiate an external deposit using the local bank's routing number

I also updated the local routing number to a random number instead of 123456789, so this issue shouldn't come up organically


kubernetes-manifests/config.yaml
```diff
@@ -17,7 +17,7 @@ kind: ConfigMap
 metadata:
   name: environment-config
 data:
-  LOCAL_ROUTING_NUM: "123456789"
+  LOCAL_ROUTING_NUM: "883745000"
```
murog, line 20: Can you update the local routing num in ui tests as well?
 https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/ad2caf03b510433111c86c83aabe7bf95f2c0e4d/.github/workflows/ui-tests/cypress.json#L49

murog: Initially thought the ui tests failed because they were flaky, but looks like `1) cannot reference local routing number` is failing

Daniel-Sanche: Good catch! I told it to run a bunch of times thinking it was just regular flakiness

PR #228, opened 2020-06-12
add logging config and log messages for contacts service

### LINKED ISSUE #213: Standardize logging - contacts

sunny3728:
This resolves #213


src/contacts/contacts.py
```diff
@@ -155,6 +162,7 @@ def _validate_new_contact(req):
 
     def _check_contact_allowed(username, accountid, req):
         """Check that this contact is allowed to be created"""
+        app.logger.debug("checking that contact is allowed to be created: %s", str(req))
```
askmeegs, line 165: nit, `s/that contact/that this contact`

kubernetes-manifests/contacts.yaml
```diff
@@ -38,6 +38,8 @@ spec:
           value: "v0.2.0"
         - name: PORT
           value: "8080"
+        - name: LOG_LEVEL
```
askmeegs, line 41: @WilliamG240  worth discussing what we want the default global vs. service log level to be.  Right now in your PR Sunny, contacts uses a global `info` and a service-level` info` -- Will, balancereader in your PR uses a global `info` and service-level `error`. Ideally all the Deployments files in `kubernetes-manifests/` should look the same -- I might actually argue that by default, *no* service has a service-level override -- the log level is commented out in the deployment, and all services use the global var.  Otherwise every service is overriding the global configmap.

src/contacts/Dockerfile
```diff
@@ -37,5 +37,8 @@ RUN pip install -r /app/requirements.txt
 # Add application code.
 ADD *.py /app/
 
+# Copy logging configuration for gunicorn
+COPY logging.conf /logging.conf
+
 # Start server using gunicorn
-CMD gunicorn -b :$PORT --threads 4 --log-level=info "contacts:create_app()"
+CMD gunicorn -b :$PORT --threads 4 --log-config /logging.conf --log-level=$LOG_LEVEL "contacts:create_app()"
```
askmeegs, line 44: re: our earlier conversation, I see what you mean now! You aren't hardcoding the log level anywhere, my mistake :)

src/contacts/contacts.py
```diff
@@ -71,11 +71,13 @@ def get_contacts(username):
                 raise PermissionError
 
             contacts_list = contacts_db.get_contacts(username)
+            app.logger.debug("Succesfully retrieved contacts.")
             return jsonify(contacts_list), 200
         except (PermissionError, jwt.exceptions.InvalidTokenError):
+            app.logger.error("Error retrieving contacts list: authentication denied")
```
askmeegs, line 77: I would actually print the `InvalidTokenError` here like you're doing with the `SQLAlchemyError` below -- and just make sure nothing specific about the incoming JWT is logged to stdout

src/contacts/contacts.py
```diff
@@ -118,17 +120,22 @@ def add_contact(username):
                 "is_external": req["is_external"],
             }
             # Add contact_data to database
+            app.logger.debug("Adding new contact to database.")
```
askmeegs, line 123: nit, s/`database`/`the database`

src/contacts/contacts.py
```diff
@@ -118,17 +120,22 @@ def add_contact(username):
...
             contacts_db.add_contact(contact_data)
+            app.logger.info("Successfully added new contact.")
             return jsonify({}), 201
 
         except (PermissionError, jwt.exceptions.InvalidTokenError):
+            app.logger.error("Error adding contact: authentication denied")
```
askmeegs, line 129: same as above, print the error

askmeegs:
@sunny3728 may be worth discussing with @WilliamG240 how to handle Java vs. Python service startup headers. The java services print block letters indicating the service name; the Python services don't. I recommend adding these to the Python services or removing them entirely. 

Example - 
https://gist.github.com/askmeegs/1e66a6a7f4a5ac8680d579d0ab7c3c57

(I'm leaning towards removing them. They look cool when running locally, but aren't formatted well in Cloud Logging.)

askmeegs:
I notice a few of the contacts endpoints have a `raise __Warning` syntax -- eg. validating a contact https://github.com/sunny3728/bank-of-anthos/blob/logging-contacts/src/contacts/contacts.py#L151 

Do these warnings get logged anywhere?

src/contacts/Dockerfile
askmeegs, line 44: +1, this makes sense!

askmeegs:
When I switched the log level to `debug` for contacts, I saw that the DB logs may not be formatted correctly -- 

```
contacts-68f845c8c4-pj6gk contacts FROM contacts
contacts-68f845c8c4-pj6gk contacts WHERE contacts.username = %(username_1)s
contacts-68f845c8c4-pj6gk contacts 2020-06-15 12:01:19 | [DEBUG] | get_contacts | RESULT: <sqlalchemy.engine.result.ResultProxy object at 0x7f1bd392aa50>
```

1. Ideally the whole DB query can fit into one log message. Eg. `FROM contacts` and `WHERE contacts..` are part of the same message (but if this is really tricky, can leave as-is)

2. not sure if `%(username_1)s` is formatted right?  should we be logging the actual username here?

sunny3728:
> I notice a few of the contacts endpoints have a `raise __Warning` syntax -- eg. validating a contact https://github.com/sunny3728/bank-of-anthos/blob/logging-contacts/src/contacts/contacts.py#L151
> 
> Do these warnings get logged anywhere?

That specific warning should get logged when it is caught in the parent function here https://github.com/sunny3728/bank-of-anthos/blob/logging-contacts/src/contacts/contacts.py#L132. In general I was just logging when errors were caught rather than when they were raised -- is that the right approach?

sunny3728:
> When I switched the log level to `debug` for contacts, I saw that the DB logs may not be formatted correctly --
> 
> ```
> contacts-68f845c8c4-pj6gk contacts FROM contacts
> contacts-68f845c8c4-pj6gk contacts WHERE contacts.username = %(username_1)s
> contacts-68f845c8c4-pj6gk contacts 2020-06-15 12:01:19 | [DEBUG] | get_contacts | RESULT: <sqlalchemy.engine.result.ResultProxy object at 0x7f1bd392aa50>
> ```
> 
> not sure if `%(username_1)s` is formatted right? should we be logging the actual username here?

I _think_ this has to do with sqlalchemy's stringifying of queries rather than the logger. Based on [this](https://docs.sqlalchemy.org/en/13/faq/sqlexpressions.html#how-do-i-render-sql-expressions-as-strings-possibly-with-bound-parameters-inlined) it seems like we could change this format. Should I?

kubernetes-manifests/contacts.yaml
Daniel-Sanche: >  I might actually argue that by default, no service has a service-level override -- the log level is commented out in the deployment, and all services use the global var. Otherwise every service is overriding the global configmap.

This is what I was going to suggest. I think we should leave service-level variables commented out on master

We also need to document the new environment variable in the service's README

src/contacts/Dockerfile
Daniel-Sanche, line 44: What happens if LOG_LEVEL isn't set? Will it cause a crash? Does it at least give a reasonable error message?

Maybe we should bake a default LOG_LEVEL value into the container to be overwritten by Kubernetes

src/contacts/contacts.py
```diff
@@ -118,17 +120,22 @@ def add_contact(username):
...
             contacts_db.add_contact(contact_data)
+            app.logger.info("Successfully added new contact.")
             return jsonify({}), 201
 
-        except (PermissionError, jwt.exceptions.InvalidTokenError):
+        except (PermissionError, jwt.exceptions.InvalidTokenError) as err:
+            app.logger.error("Error adding contact: %s", str(err))
             return "authentication denied", 401
         except UserWarning as warn:
+            app.logger.error("Error adding contact: %s", str(warn))
```
djmailhot, line 132: This logging level should not be `error`

This situation is invalid user input.  The application is operating normally when it throws this exception.

Instead it should be either `warn` (if you believe users are always perfect at following rules and this is unexpected behavior), or it should be `info` (because it would be nice to know that the contact could not be added).  Again, you can update the log message to reflect this:  e.g. "Failed to add contact:  invalid input"

src/contacts/contacts.py
```diff
@@ -71,11 +71,13 @@ def get_contacts(username):
                 raise PermissionError
 
             contacts_list = contacts_db.get_contacts(username)
+            app.logger.debug("Succesfully retrieved contacts.")
             return jsonify(contacts_list), 200
-        except (PermissionError, jwt.exceptions.InvalidTokenError):
+        except (PermissionError, jwt.exceptions.InvalidTokenError) as err:
+            app.logger.error("Error retrieving contacts list: %s", str(err))
```
djmailhot, line 77: This logging level should not be `error`

This situation is catching an incorrect login token.  This is the fault of the user, and the application is actually operating exactly as it is supposed to when it raises this error.

Instead it should be either `warn` (if you believe this is unexpected behavior) or `info` (because while the application is actually operating as expected when refusing incorrect authentication, it would be nice to know that the contacts could not be retrieved).  ALSO, update the log message to reflect the new log level (don't use the word 'Error'):  e.g. "Failed to retrieve contacts list:  invalid permissions"

src/contacts/Dockerfile
sunny3728: It seems like gunicorn sets the log level to info by [default](https://docs.gunicorn.org/en/stable/settings.html#loglevel) if the flag isn't specified. With `--log-level=$LOG_LEVEL`, if I don't specify the LOG_LEVEL anywhere, it seems like it also defaults to `info`.

kubernetes-manifests/contacts.yaml
```diff
@@ -28,7 +28,7 @@ spec:
       terminationGracePeriodSeconds: 5
       containers:
       - name: contacts
-        image: gcr.io/bank-of-anthos/contacts:latest
+        image: gcr.io/sunnyqi/contacts:0616201501
```
WilliamG240, line 31: Seems like this should stay bank-of-anthos?

sunny3728: oh right! didn't mean to commit that

kubernetes-manifests/contacts.yaml
sunny3728: For some reason the contacts service [README](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/src/contacts/README.md) already mentions the LOG_LEVEL env var, though I don't think it was used anywhere before?

askmeegs: > > When I switched the log level to `debug` for contacts, I saw that the DB logs may not be formatted correctly --
> > ```
> > contacts-68f845c8c4-pj6gk contacts FROM contacts
> > contacts-68f845c8c4-pj6gk contacts WHERE contacts.username = %(username_1)s
> > contacts-68f845c8c4-pj6gk contacts 2020-06-15 12:01:19 | [DEBUG] | get_contacts | RESULT: <sqlalchemy.engine.result.ResultProxy object at 0x7f1bd392aa50>
> > ```
> > 
> > 
> > not sure if `%(username_1)s` is formatted right? should we be logging the actual username here?
> 
> I _think_ this has to do with sqlalchemy's stringifying of queries rather than the logger. Based on [this](https://docs.sqlalchemy.org/en/13/faq/sqlexpressions.html#rendering-bound-parameters-inline) it seems like we could change this format, but it introduces some security issues and isn't recommended. Should I still change this?

nope you can keep it as is!

Daniel-Sanche:
I'm having trouble changing the log levels. It seems to always show the [INFO] logs, no matter what variable is set in the container. Do you know what might be causing that?

sunny3728:
> I'm having trouble changing the log levels. It seems to always show the [INFO] logs, no matter what variable is set in the container. Do you know what might be causing that?

Hm, I'm not sure, I was able to change it to `debug` and `error` and see the right logs, I believe. What were you setting the level to?

Daniel-Sanche:
Ok, I was able to reproduce my issue:
1. clear out the cluster with `kubectl delete configmaps --all; kubectl delete deployment --all`
2. comment out LOG_LEVEL in contacts, set it to debug in the configmap
3. deploy with `make deploy PROJECT_ID=...`
4. uncomment contacts LOG_LEVEL and set to `error`
5. `kubectl apply -f contacts.yaml`
6. view contacts logs

I'd expect the new contacts pod to display error logs, but it's showing info (which is neither setting). I can exec into the pod and see that the variable is set properly. I don't quite know what to make of this yet, but it seems like Kubernetes isn't injecting its variables until after gunicorn has already started.

Let me know if you're able to see the same issue. I'll try to help investigate more soon

Daniel-Sanche:
The config map has this comment:

`# valid values of log level: fatal,error,warn,info,debug`

`warn` doesn't work for the Python logger though, right? Are there any other incompatibilities between languages?

sunny3728:
> The config map has this comment:
> 
> `# valid values of log level: fatal,error,warn,info,debug`
> 
> `warn` doesn't work for the Python logger though, right? Are there any other incompatibilities between languages?

Log4j2 also uses `fatal` while python uses `critical`.

Tracking this issue with #239

PR #246, opened 2020-06-18
Default latest transaction id set to -1 in TransactionRepository

### LINKED ISSUE #238: App fails to deploy if transaction database is empty
The balance reader service is unable to correctly instantiate when there are no transactions in the database.  It hits an issue with retrieving TransactionRepository.latestId() and getting a Null value.

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'balanceReaderController' defined in URL [jar:file:/opt/monolith/ledgermonolith.jar!/BOOT-IN
F/classes!/anthos/samples/financedemo/ledgermonolith/BalanceReaderController.class]: Bean instantiation via constructor failed; nested exception is org.springframework.beans.BeanI
nstantiationException: Failed to instantiate [anthos.samples.financedemo.ledgermonolith.BalanceReaderController]: Constructor threw exception; nested exception is org.springframew
ork.aop.AopInvocationException: Null return value from advice does not match primitive return type for: public abstract long anthos.samples.financedemo.ledgermonolith.TransactionR
epository.latestId()
```

djmailhot:
Resolves #238 

Alternative approach is to do it directly in the Java code of each Controller class.


src/balancereader/src/main/java/anthos/samples/financedemo/balancereader/TransactionRepository.java
```diff
@@ -40,6 +40,12 @@
...
-    @Query("SELECT MAX(transactionId) FROM Transaction")
-    long latestId();
+    /**
+     * Returns the id of the latest transaction, or -1 if no transactions exist.
+     */
+    @Query(value = "SELECT COALESCE("
+        + "(SELECT MAX(transaction_id) FROM transactions), "
+        + "CAST ('-1' AS BIGINT))",
+        nativeQuery = true)
+    Long latestId();
```
Daniel-Sanche, line 50: was there a reason for going from long to Long?

djmailhot: Long (instead of long) allowed the query to return NULL.  Not as important anymore since I defaulted the query to return -1 by default.

Daniel-Sanche: Would we have to add any null checks to other parts of the code?

djmailhot: We don't have to now since this query returns a default value.

If you think it makes more sense to do NULL checks in the code, I can do that instead of fixing the query.

Daniel-Sanche: I guess my question is: if we're using a default value and don't expect nulls, why not keep using long?

djmailhot: Using `Long` feels less brittle to me.  It protects against this failure mode from any future changes to the query that may make it return Null again.  'Defense in depth' if you will.

I sense there is resistance to having it `Long`.  I'm assuming that you're concerned about performance efficiency.

Daniel-Sanche: Definitely not worried about performance. But the LedgerReader class is still expecting long instead of Long, so if you make this change here, you should also make it there. And also add null checks there so it fails in an expected way if it does end up being null. (I'm not really sure what the error would look like if the Long was null right now)

Daniel-Sanche: Basically, if we're confident that query always returns a value (I didn't read the docs, but it seems like it should), I think it's safe to keep using long. If there's a chance it still returns null, we need to change other code to handle that error case.

djmailhot: > But the LedgerReader class is still expecting long instead of Long, so if you make this change here, you should also make it there. 

This actually isn't an issue, see "Java autoboxing".

> And also add null checks there so it fails in an expected way if it does end up being null. (I'm not really sure what the error would look like if the Long was null right now)

That is a fair point, if I'm protecting against Null here in TransactionRepository, then I also still need to do it in LedgerReader.

Okay, you've convinced me to just do the check in the Java code.  I've removed the overly complicated SQL query and added null checks in the Java code.

Makefile
```diff
@@ -14,7 +14,6 @@
 
 .-PHONY: cluster deploy deploy-continuous logs checkstyle check-env
 
-ZONE=us-west1-a
```
Daniel-Sanche, line 17: is there a reason for removing this?

kubernetes-manifests/config.yaml
```diff
@@ -36,7 +36,7 @@ kind: ConfigMap
 metadata:
   name: demo-data-config
 data:
-  USE_DEMO_DATA: "True"
+  USE_DEMO_DATA: "False"
```
Daniel-Sanche, line 39: I assume this should be reverted before merging?

src/balancereader/src/main/java/anthos/samples/financedemo/balancereader/TransactionRepository.java
Daniel-Sanche: > This actually isn't an issue, see "Java autoboxing".

It looks like Autoboxing thows a NullPointerException if the Long is null though, and allowing nulls was the argument for using Long here in the first place

> Okay, you've convinced me to just do the check in the Java code. I've removed the overly complicated SQL query and added null checks in the Java code.

Ok cool, that sounds like a good solution to me, I'll re-review

Makefile
djmailhot: Yes, but actually no.

Accidentally included it in the commit

kubernetes-manifests/config.yaml
djmailhot: Yep, accidentally included it in the commit

Daniel-Sanche:
Looks like you're still working, let me know when you're ready for a re-review

djmailhot:
> Looks like you're still working, let me know when you're ready for a re-review

Ready for review

src/balancereader/src/main/java/anthos/samples/bankofanthos/balancereader/LedgerReader.java
```diff
@@ -73,7 +74,8 @@ public void startWithCallback(LedgerReaderCallback callback)
         this.callback = callback;
         // get the latest transaction id in ledger
         try {
-            this.latestId = dbRepo.latestId();
+            Long dbId = dbRepo.latestId();
+            this.latestId = (dbId != null ? dbId : this.latestId);
```
Daniel-Sanche, line 78: I personally find these a bit hard to read. 

I'd say the could be done cleaner with a null check, since the else case is a no-op:
```
if (dbId != null){
   this.latestId = dbId
}
```

It's also a bit hard to read with 3 similarly named variables interacting with each other. It might be better if you could remove the temporary `dbId` variable (but this would require some changes to handle/remove nulls later)
```
this.latestId = dbRepo.latestId();
```

Or maybe just some extra comments would help

But these are just style issues, so feel free to ignore

src/transactionhistory/src/main/java/anthos/samples/bankofanthos/transactionhistory/LedgerReader.java
```diff
@@ -71,9 +72,11 @@ public void startWithCallback(LedgerReaderCallback callback)
             throw new IllegalStateException("callback is null");
         }
         this.callback = callback;
+        this.latestId = STARTING_TRANSACTION_ID;
```
Daniel-Sanche, line 75: Looks like this change didn't make it over to the other LedgerReader. latestId was left uninitialized there

djmailhot: Good catch.  Done.

src/balancereader/src/main/java/anthos/samples/bankofanthos/balancereader/LedgerReader.java
djmailhot: Roger that.  I agree, it is hard to read and understand.  So I made some readability changes that I think makes the code easier to understand.

PR #241, opened 2020-06-18
adds standardized logging for ledgerwriter service

WilliamG240:
This fixes #215.

Example log:
image

PR #248, opened 2020-06-19
removes global log level env var

### LINKED ISSUE #239: remove or resolve inconsistency for global log level env variable
The python (gunicorn) logger uses [these](https://docs.gunicorn.org/en/stable/settings.html#loglevel) levels while the java services logger (log4j2) uses [these](https://logging.apache.org/log4j/2.x/manual/customloglevels.html) which has discrepancies (warn/warning and fatal/critical).

We need a way to work around this inconsistency between the services log level, in order to use a global log level variable.

Possible solutions:
- Remove the global configmap variable and have each service default to the same log level
- Use one global variable for java services and a different one for python services
- parse the log levels in the service to handle the inconsistencies (not recommended)
- find different logging libraries that use consistent log levels

WilliamG240:
This fixes #239

### DIFF
--- a/kubernetes-manifests/balance-reader.yaml
+++ b/kubernetes-manifests/balance-reader.yaml
@@ -45,9 +45,10 @@ spec:
         # tell Java to obey container memory limits
         - name: JVM_OPTS
           value: "-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap"
-        # service level override of log level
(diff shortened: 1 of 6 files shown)


WilliamG240:
> What about `ledger-writer.yaml`?

Has already been removed in #241!

djmailhot:
> > What about `ledger-writer.yaml`?
> 
> Has already been removed in #241!

Roger roger!

PR #249, opened 2020-06-19
sets default log level for Java services

### LINKED ISSUE #247: Set a default log level in java services
All environment variables in java services are currently required, and if they aren't set, the service exits. LOG_LEVEL should be an exception to this rule, and should default to INFO if no log level is set in the manifests.

WilliamG240:
This fixes #247

### DIFF
--- a/.github/workflows/ui-tests/cypress.json
+++ b/.github/workflows/ui-tests/cypress.json
@@ -46,7 +46,7 @@
                     "name": "Bob"
                 }
             ],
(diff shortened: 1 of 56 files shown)


WilliamG240:
> What about the LedgerWriter?

It has already been preemptively added to the Ledgerwriter in #241

djmailhot:
> > What about the LedgerWriter?
> 
> It has already been preemptively added to the Ledgerwriter in #241

Roger roger!

PR #259, opened 2020-06-30
add tracing instrumentation for java services

### LINKED ISSUE #203: Add tracing for ledger writer

WilliamG240:
This fixes #203, #204, #205 
Example trace for Java services:
image


src/ledgerwriter/src/main/java/anthos/samples/bankofanthos/ledgerwriter/LedgerWriterController.java
```diff
@@ -72,6 +73,9 @@
     public static final String UNAUTHORIZED_CODE = "not authorized";
     public static final String JWT_ACCOUNT_KEY = "acct";
 
+    @Autowired
```
djmailhot, line 76: To be clear, is this change required to enable tracing?

WilliamG240: Yup it's to enable tracing, the restTemplate needs to be autowired to allow Spring Cloud sleuth to inject an Interceptor to the request.

WilliamG240: Also the fix regarding paging was because in order to enable tracing I had to update the spring boot version and the PageRequest was deprecated which is why I had to update the function call.

djmailhot: Thanks for the explanation.  Everything sounds good.

PR #260, opened 2020-06-30
Remove default ZONE from Makefile to allow specifying ZONE

djmailhot:


Daniel-Sanche:
Why remove the default? You can always run `make cluster PROJECT_ID=custom ZONE=us-central1-a` to override

djmailhot:
> Why remove the default? You can always run `make cluster PROJECT_ID=custom ZONE=us-central1-a` to override

I was not aware that was an option...!

Thanks, closing this PR

PR #263, opened 2020-07-08
Fix Deposit UI Test

Daniel-Sanche:
We had a flaky UI test. Sometimes it would be looking for a value $0.01 off from the true value. This likely comes down to float math issues.

This PR aims to keep stuff simple by just comparing dollar values


Daniel-Sanche:
This passed CI once. But I'm going to try running a couple times to make sure the flakiness is gone

murog:
Thanks for adding this fix Dan!  Can you update transfer_spec as well?  `see balance update after transfer` test is still flaky

https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/cb1d46e9aa90af4800d8083a047766ea68c33c35/.github/workflows/ui-tests/cypress/integration/transfer_spec.js#L99

Daniel-Sanche:
Good catch, I added the fix there as well

PR #264, opened 2020-07-12
Accounts Service - combines userservice, contacts

askmeegs:


askmeegs:
@googlebot I fixed it.

PR #268, opened 2020-07-16
Fixes loadgenerator not finishing signup

### LINKED ISSUE #267: Loadgenerator never finishing signup
After the merging of #172, the loadgenerator is no longer finishing signup due to the creation of the uuid with is longer than 15 characters and contains dashes.

WilliamG240:
This PR fixes #267, this uses a random generation of 15 character alphanumeric string in order to satisfy username constraints specified by #172. 

The downside of replacing the uuid creation is that we now have a probability of collisions, but since we have 62 total alphanumeric characters and a 15 character string, theres a 1/62^15 chance of collision and the number of usernames the `loadgenerator` would need to create to even get a 1% probability of collisions is > 0.01* 62^15 = 7.68*10^24. Also increased backend timeout since the userservice would finish the user creation but the frontend timeout would end just after.

Before:
image

After:
image

### DIFF
--- a/src/frontend/flask_server.py
+++ b/src/frontend/flask_server.py
@@ -496,7 +496,7 @@ APP.config["CONTACTS_URI"] = 'http://{}/contacts'.format(
     os.environ.get('CONTACTS_API_ADDR'))
 APP.config['PUBLIC_KEY'] = open(os.environ.get('PUB_KEY_PATH'), 'r').read()
 APP.config['LOCAL_ROUTING'] = os.getenv('LOCAL_ROUTING_NUM')
-APP.config['BACKEND_TIMEOUT'] = 3  # timeout in seconds for calls to the backend
+APP.config['BACKEND_TIMEOUT'] = 4  # timeout in seconds for calls to the backend
 APP.config['TOKEN_NAME'] = 'token'
 APP.config['TIMESTAMP_FORMAT'] = '%Y-%m-%dT%H:%M:%S.%f%z'
 APP.config['SCHEME'] = os.environ.get('SCHEME', 'http')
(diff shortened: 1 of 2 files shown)


src/loadgenerator/locustfile.py
```diff
@@ -57,6 +57,12 @@ def signup_helper(locust, username):
             response.failure("login failed")
         return found_token
 
+def generate_username():
+    """
+    generates random 15 character
+    alphanumeric username
+    """
+    return ''.join(random.choice(ascii_letters + digits) for _ in range(15))
```
murog, line 65: Nice! Very clean

src/frontend/flask_server.py
```diff
@@ -496,7 +496,7 @@ def format_currency(int_amount):
     os.environ.get('CONTACTS_API_ADDR'))
 APP.config['PUBLIC_KEY'] = open(os.environ.get('PUB_KEY_PATH'), 'r').read()
 APP.config['LOCAL_ROUTING'] = os.getenv('LOCAL_ROUTING_NUM')
-APP.config['BACKEND_TIMEOUT'] = 3  # timeout in seconds for calls to the backend
+APP.config['BACKEND_TIMEOUT'] = 4  # timeout in seconds for calls to the backend
```
murog, line 499: Was the timeout increase necessary because of the generate username changes?  Or was it needed before this change?

WilliamG240: It seems that it is necessary after the generating of username change, I think it is the overhead of the for loop taking a little longer (couple milliseconds) more than expected.

murog: sg, thanks for clearing that up.  LGTM

PR #285, opened 2020-07-21
Trace-log correlation java testing

WilliamG240:
This implements trace-log correlation for the Java services. In order to integrate trace-log correlation the following changes were made:
- Swap of `log4j2` logging framework with `logback`, this is due to the fact that Spring Cloud Sleuth is currently missing integration with `log4j2` in regards to associating a log with trace ID and span ID.
   - As a consequence of this swap the FATAL log level has to be removed as logback does not support it.
- Added extra log output type with environment variable for JSON log output for trace-log correlation in Stackdriver (does not intake txt logs for trace-log correlation) 

Areas of development:
- Look into ways for log4j2 to associate traceID and spanID
- Alternatives to removing the FATAL log level (Work with sl4j2 [markers](http://www.slf4j.org/apidocs/org/slf4j/Marker.html)) 
- Fluentd customization for the removal of the 2 log output format, and keep only the console txt format


askmeegs:
Hi @WilliamG240 - closing this draft PR since it will not merge imminently. You can still use the closed PR for your own records + safe keeping :)

PR #290, opened 2020-07-27
add jacoco coverage reports and github action

sunny3728:


.github/workflows/ci.yaml
```diff
@@ -66,6 +66,24 @@ jobs:
...
+      timeout-minutes: 20
+      run: |
+        set -x
+        for SERVICE in "balancereader" "ledgerwriter"; do
+          echo "checking $SERVICE..."
+          # save current working dir to memory and cd to src/$SERVICE
+          pushd src/$SERVICE
+            mvn jacoco:report
+            echo "Coverage for $SERVICE:"
+            awk -F, \
```
sunny3728, line 79: JaCoCo doesn't have a built-in way to show coverage in the console, so I am using a workaround suggested [here](https://stackoverflow.com/questions/49919512/jacoco-coverage-percentage-to-print-on-console)

.github/workflows/ci.yaml
```diff
@@ -66,6 +66,24 @@ jobs:
           # return to previously saved path
           popd
         done
+    - name: Java Code Coverage
+      timeout-minutes: 20
+      run: |
+        set -x
```
murog, line 72: Thanks for finding this workaround!  Would it be possible to make the output more readable, currently prints ~80 lines so it is difficult to see where the coverage is

askmeegs: +1

sunny3728: Is it possible to prevent github actions from printing out the commands it is running? That seems to be most of it

sunny3728: Whoops, my bad, this was just happening because I left the "set -x" in there!

PR #291, opened 2020-07-30
add github action for python code coverage

sunny3728:


.github/workflows/ci.yaml
```diff
@@ -83,6 +83,21 @@ jobs:
           # return to previously saved path
           popd
         done
+    - name: Python Code Coverage
```
sunny3728, line 86: I'm basically repeating everything from the unit tests step, but running the code coverage command instead. Would it make sense to combine these steps and just print the code coverage after running the tests, or should this be a separate action in case we add minimum coverage checks later?

Issue #297, opened 2020-08-03
Jacoco plugin breaks log4j in the Java services


askmeegs:
```
Exception in thread "main" java.lang.IllegalArgumentException: LoggerFactory is not a Logback LoggerContext but Logback is on the classpath. Either remove Logback or the competing implementation (class org.apache.logging.slf4j.Log4jLoggerFactory loaded from file:/app/libs/log4j-slf4j-impl-2.13.3.jar). If you are using WebLogic you will need to add 'org.slf4j' to prefer-application-packages in WEB-INF/weblogic.xml: org.apache.logging.slf4j.Log4jLoggerFactory
	at org.springframework.util.Assert.instanceCheckFailed(Assert.java:696)
	at org.springframework.util.Assert.isInstanceOf(Assert.java:596)
	at org.springframework.boot.logging.logback.LogbackLoggingSystem.getLoggerContext(LogbackLoggingSystem.java:284)
	at org.springframework.boot.logging.logback.LogbackLoggingSystem.beforeInitialize(LogbackLoggingSystem.java:104)
	at org.springframework.boot.context.logging.LoggingApplicationListener.onApplicationStartingEvent(LoggingApplicationListener.java:232)
	at org.springframework.boot.context.logging.LoggingApplicationListener.onApplicationEvent(LoggingApplicationListener.java:213)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener(SimpleApplicationEventMulticaster.java:172)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:165)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:139)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:127)
	at org.springframework.boot.context.event.EventPublishingRunListener.starting(EventPublishingRunListener.java:74)
```

stevemcghee:
im still seeing this failure on a fresh turnup of boa

Issue #270, opened 2020-07-18
CI Migration: Move integration testing from Kind to GKE


murog:
## Description
Current integration tests use an ephemeral Kind cluster that is hosted on a GCE instance.  Moving to GKE would allow us to test future Anthos integrations.

## Expected Behavior
- [x] Create Test GKE cluster
- [ ] Update [ci.yaml](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/.github/workflows/ci.yaml) to:
    - [x]  authenticate to gke test cluster
    - [x]  create test namespace
    - [x] deploy to test namespace
    - [ ] get external ip of deployment
    - [ ]  run integration tests against deployment

PR #299, opened 2020-08-03
PR deploy tests: Kind -> GKE, fix skaffold errs

### LINKED ISSUE #298: Github actions deploy tests use cached images
It looks like the Github actions deployment tests build from the local branch - but also rely on cached images (skaffold defaults to using a local image cache) -- meaning that what gets deployed may be stale / not what's actually in the PR. This may be causing false positives in the deploy tests.

askmeegs:
Fixes #298, #270 

**Note** - deploy tests are currently failing *as expected*, given the ongoing Java logger dependency bug - #297 

- [x] updates deploy tests to use an existing PR cluster (GKE) instead of spinning up a Kind cluster on every build. This is a *separate* GKE cluster from the master cluster. `push-and-deploy` CI for the master branch is unchanged.  
- [x] skaffold now builds dedicated images per PR # - [looks like this in GCR](https://pantheon.corp.google.com/gcr/images/bank-of-anthos/GLOBAL/refs/pull?gcrImageListsize=30)
- [x] each PR has its own deploy namespace in the GKE cluster that spins up on every build. When there are 2 PRs both using the cluster, the namespaces in the cluster look like this:  

```
pr299             Active   3m28s
pr303             Active   105s
```

**Note** - we may need to set up autoscaling on the cluster (or spin up a few more test clusters) if we hit resource issues. Cluster only has 4 nodes.

- [x] verified that skaffold is in fact building + deploying the right images for a given PR. :)  (tested this by adding a distinct log message to 2 different PRs, and checking pod logs during the build tests.) 
- [x] added cleanup step that runs *no matter if the deploy tests succeed or fail* - which deletes the PR-specific namespace.
- [x] reduce deploy test pod timeout from 8m30s to 5m (`300` seconds) - we should fail any pod that takes more than 5 minutes to start :)


.github/workflows/ci.yaml
```diff
@@ -101,38 +101,51 @@ jobs:
...
-        kubectl wait --for=condition=available --timeout=500s deployment/userservice
-        kubectl wait --for=condition=available --timeout=500s deployment/ledgerwriter
-        kubectl wait --for=condition=available --timeout=500s deployment/transactionhistory
-        kubectl wait --for=condition=available --timeout=500s deployment/balancereader
-        kubectl wait --for=condition=available --timeout=500s deployment/loadgenerator
-        kubectl wait --for=condition=ready --timeout=500s pod/accounts-db-0
-        kubectl wait --for=condition=ready --timeout=500s pod/ledger-db-0
+        PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
+        NAMESPACE="pr${PR_NUMBER}"
+        kubectl config set-context --current --namespace=$NAMESPACE
```
murog, line 133: appreciate the shorter timeout (-:

.github/workflows/ci.yaml
```diff
@@ -101,38 +101,51 @@ jobs:
...
+        kubectl wait --for=condition=available --timeout=300s deployment/frontend
+        kubectl wait --for=condition=available --timeout=300s deployment/contacts
+        kubectl wait --for=condition=available --timeout=300s deployment/userservice
+        kubectl wait --for=condition=available --timeout=300s deployment/ledgerwriter
+        kubectl wait --for=condition=available --timeout=300s deployment/transactionhistory
+        kubectl wait --for=condition=available --timeout=300s deployment/balancereader
+        kubectl wait --for=condition=available --timeout=300s deployment/loadgenerator
+        kubectl wait --for=condition=ready --timeout=300s pod/accounts-db-0
+        kubectl wait --for=condition=ready --timeout=300s pod/ledger-db-0
     - name: Port Forward frontend
```
murog, line 143: Do we still need to port forward the frontend?  Now that we're on a GKE cluster, can we use the live IP address that is exposed by the ingress instead?  Let me know if you think that's out of scope for this PR and I can open an issue

.github/workflows/ci.yaml
```diff
@@ -101,38 +101,51 @@ jobs:
...
-        skaffold config set --global local-cluster true
-        skaffold run --default-repo local \
-                     --profile ${{ matrix.profile }}
+        PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
+        NAMESPACE="pr${PR_NUMBER}"
+        gcloud container clusters get-credentials $PR_CLUSTER --zone $ZONE --project $PROJECT_ID
+        kubectl create namespace $NAMESPACE
+        skaffold config set --global local-cluster false
+        skaffold run --default-repo=gcr.io/$PROJECT_ID/$GITHUB_REF --tag=$GITHUB_SHA --namespace=$NAMESPACE
+      env:
```
murog, line 123: can we have an env for the entire workflow instead?  WDYT?

.github/workflows/ci.yaml
```diff
@@ -142,6 +155,19 @@ jobs:
     - name: Ping frontend
       run: |
           curl --silent --output /dev/null --show-error --fail localhost:3000
-    - name: UI tests
+    - name: Run UI tests (Cypress)
       run: |
           docker run -t -v $PWD/.github/workflows/ui-tests:/e2e -w /e2e --network=host cypress/included:4.3.0
```
murog, line 160: If we used the external ip address instead this would get updated to include `-e CYPRESS_baseURL=${EXTERNAL_IP}`

.github/workflows/ci.yaml
askmeegs: Will leave for now / to update in @WilliamG240's staging work

askmeegs: I think this is in scope, updating!

.github/workflows/ci.yaml
askmeegs: Thanks! updated.

murog:
re-running ci

askmeegs:
@murog it looks like the external LB IP for the frontend is flaky (sometimes throws a `connection refused` on ping) and also might cause CORS errors in the UI tests (?). I removed the external IP for now, and re-added the local pod port-forwarding. 

For staging purposes, we'll need to debug this. filing an issue.

Issue #294, opened 2020-08-02
Transaction history labels debits as credits and credits as debits


salsferrazza:
Debits from an account are typically shown as subtractions with a corresponding negative figure (e.g. `-12.42` or `(12.42)` if you are an accountant) and credits to an account with a positive figure (e.g. `+12.42` or simply `12.42`).

Similarly, the color coding would be expected as being green for Credits and red for Debits. In the current build, the color coding is indeed correct but the label of "Credit" or "Debit" is reversed.

![image](https://user-images.githubusercontent.com/5541617/89127115-1be12680-d4b9-11ea-89d7-9473d6c58c8b.png)

askmeegs:
Thanks for this suggestion, however I'm seeing in my own (real) bank account, for instance, that debits ($ flowing into the account) show up in green, with `+ $XX` and credits show up in gray, as `- $XX`. This is consistent with the Bank of Anthos UI.  

![image](https://user-images.githubusercontent.com/3137106/105901970-040fd880-5fec-11eb-943d-9d2056c9a724.png)


And I think we are using these terms correctly based on this -- https://www.accountingtools.com/articles/2017/5/17/debits-and-credits  (Debits = $ in, Credit = $ out) - in this case it's not debit "from" the Bank of Anthos account, it's debit "to." ie. a deposit into the user's account. 

Closing this issue for now.

mservidio:
Hi @askmeegs - For the example, who is the 'user' viewing the transactions? Is it the bank customer?

PR #302, opened 2020-08-03
Swap labels on credits and debits

salsferrazza:
copy swap for credit/debit labels as per #294


askmeegs:
Addressed separately in #437

PR #354, opened 2020-09-17
Add initial troubleshooting guide

### LINKED ISSUE #322: Write troubleshooting guide

askmeegs:
Fixes #322

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -40,6 +40,15 @@ transactionhistory   ClusterIP      10.48.20.206   <none>          8080/TCP     
 userservice          ClusterIP      10.48.19.11    <none>          8080/TCP       23d
 ```
 
+- The next step to verify your issue is to **clean deploy** the Bank of Anthos app to your cluster:
+
+```
+kubectl delete -f kubernetes-manifests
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -0,0 +1,95 @@
...
+```
+Events:
+  Type     Reason            Age                From               Message
+  ----     ------            ----               ----               -------
+  Warning  FailedScheduling  12s (x3 over 13s)  default-scheduler  0/1 nodes are available: 1 Insufficient memory.
+```
+
+This means that your cluster's [Node Pools](https://cloud.google.com/kubernetes-engine/docs/concepts/node-pools) do not have enough capacity to host all the Bank of Anthos workloads, and you either need to use a different cluster, or increase your existing cluster's capacity (see the [GKE docs](https://cloud.google.com/kubernetes-engine/docs/how-to/resizing-a-cluster)). 
+
+### `503` error in the frontend
```
murog, line 81: good call out

murog (approved):
I wonder if it would be helpful to explicitly call out how to reset the entire deployment? WDYT? Otherwise, LGTM!

Issue #295, opened 2020-08-03
[Tracking] Contacts service has > 1s latency


askmeegs:
<img width="684" alt="Screen Shot 2020-08-03 at 1 09 52 PM" src="https://user-images.githubusercontent.com/3137106/89208597-a6458b00-d58a-11ea-9d24-c519c064af06.png">

djmailhot:
After fiddling with this for a long while, I'm inclined to believe **this unexpected latency is common to all our Python services**.  

I inserted HTTP calls to the /ready endpoint for the contactservice and the userservice, and both calls have large latencies even though the function is simply  `return "ok", 200`.
![frontend_trace_ready](https://user-images.githubusercontent.com/704680/93639415-cd76e100-f9ad-11ea-8f47-cc46e7876579.png)

askmeegs:
Next steps: rule out - 

1. flask
2. gunicorn
3. running locally vs. in k8s

djmailhot:
I've think I've found one other source of latency:  SQLAlchemy.

It turns out that [the SQLAlchemy ORM introduces overhead latency](https://anonbadger.wordpress.com/2010/09/09/optimizing-an-sqlalchemy-call/).  This can be avoided by using raw SQL calls instead.

The latency difference between raw SQL and SQL Alchemy seems to be around ~500ms
![frontend_latency_sqlalchemy](https://user-images.githubusercontent.com/704680/93649600-361b8900-f9c1-11ea-91f4-486652e7fef0.png)

The traces seem to indicate that this is contributing to the latency problem for the contacts service.

Trace using SQL Alchemy.
![sqlalchemy_trace](https://user-images.githubusercontent.com/704680/93649831-bb06a280-f9c1-11ea-96f9-5a5d5a7303e8.png)

Trace using raw SQL
![sql_trace](https://user-images.githubusercontent.com/704680/93649838-bf32c000-f9c1-11ea-9765-878a54e7b907.png)

The code for raw sql can be found under the GitHub branch `python-raw-sql`

djmailhot:
## So far ruled out the following:

### SQL latency
**Hypothesis**:  SQL calls from the Python services are slow.  Calls not using SQL should be fast.

- Insert queries to `/ready` endpoints from the homepage (See branch `frontend-test-ready`)

**Result**:  Calls to `/ready` still have average ~500ms latencies.

### Threading & Scaling issues
[...]

PR #360, opened 2020-09-19
Upgrade OpenTelemetry (Python) + BatchExportSpanProcessor 

### LINKED ISSUE #356: All Python services have very slow HTTP requests
After fiddling with this for a long while, I'm inclined to believe **this unexpected latency is common to all our Python services**.  

I inserted HTTP calls to the /ready endpoint for the contactservice and the userservice, and both calls have large latencies even though the function is simply  `return "ok", 200`.
frontend_trace_ready

_Originally posted by @djmailhot in https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/295#issuecomment-695057813_

askmeegs:
Fixes #356 and #295. 

- [x] upgrade OpenTelemetry from `v0.10b0` to `v0.13b0`  (latest stable) 
- [x] for all 3 python services, switched the span export method from `Simple` to `Batch` - **this was causing the latency bug**. 
- [x] removed `requirements-test.txt` for the contacts and userservice tests. Now they rely on the service's actual `requirements.txt` file.  
- [x] turned off trace export `(ENABLE_TRACING-false)` in the unit tests, since setting to `TRUE` would test an integration. ideally in the future we mock (or find a mock for) the OT Cloud Trace library and add that to the tests instead.   

Now most calls to `/home` look like this (vs. 6-8 seconds long) 

Screen Shot 2020-09-23 at 12 26 48 PM


src/contacts/pyvenv.cfg
```diff
@@ -0,0 +1,3 @@
+home = /usr/local/bin
+include-system-site-packages = false
```
murog, line 2: was this file generated?  wondering if this is related to the dep mismatch?

src/contacts/tests/test_contacts.py
```diff
@@ -59,7 +59,7 @@ def setUp(self):
                     "VERSION": "1",
                     "LOCAL_ROUTING": "123456789",
                     "PUBLIC_KEY": "1",
-                    "ENABLE_TRACING": "true",
+                    "ENABLE_TRACING": "false",
```
murog, line 62: 👍

src/userservice/userservice.py
```diff
@@ -228,10 +231,10 @@ def _shutdown():
         trace.set_tracer_provider(TracerProvider())
         cloud_trace_exporter = CloudTraceSpanExporter()
         trace.get_tracer_provider().add_span_processor(
-            SimpleExportSpanProcessor(cloud_trace_exporter)
+            BatchExportSpanProcessor(cloud_trace_exporter)
         )
-        set_global_httptextformat(CloudTraceFormatPropagator())
-        # Add Flask auto-instrumentation for tracing
+        # set_global_httptextformat(CloudTraceFormatPropagator())
```
murog, line 236: why is this commented out here and in contacts.py?

murog (commented):
added a couple questions

murog (approved):

.github/workflows/ci-master.yaml
```diff
@@ -154,6 +154,7 @@ jobs:
         done
         EXTERNAL_IP=$(get_externalIP)
         echo "::set-env name=EXTERNAL_IP::$EXTERNAL_IP"
+        sleep 60
```
murog, line 157: issue #368

PR #389, opened 2020-11-05
postgres:12-alpine

mathieu-benoit:
- Update `accounts-db` and `ledger-db` container base images from `postgres:12` to `postgres:12-alpine`.
- `accounts-db`: from 109.05 MB to 59.15 MB
- `ledger-db`: from 109.05 MB to 59.14 MB

No issues with the current version but that's a way to get smaller as well as more secure (less surface) images for the databases. FYI: that's also what was done with the `redis` container on that repo: https://github.com/GoogleCloudPlatform/microservices-demo/blob/master/kubernetes-manifests/redis.yaml#L30.


mathieu-benoit:
Ready for review and comment, thanks!

askmeegs (approved):

PR #394, opened 2020-11-13
Enable tracing for ledgermonolith services via k8 manifests

### LINKED ISSUE #393: Ledgermonolith Key error for 'ENABLE_TRACING' using 'skaffold run'
When running “skaffold run…” command there is a Key error for the “frontend”, “userservice” and “contactservice” related to the “ENABLE_TRACING” environment variable.

djmailhot:
Fixes #393  .

Change summary:
- Update ledgermonolith k8 manifests with ENABLE_TRACING = 'True'

### DIFF
--- a/src/ledgermonolith/kubernetes-manifests/contacts.yaml
+++ b/src/ledgermonolith/kubernetes-manifests/contacts.yaml
@@ -38,6 +38,8 @@ spec:
           value: "v0.2.0"
         - name: PORT
           value: "8080"
+        - name: ENABLE_TRACING
+          value: "true"
         # Valid levels are debug, info, warning, error, critical.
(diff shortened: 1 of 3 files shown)


askmeegs:
The enable tracing env var handling had to be built into the code. And it's predicated on OpenTelemetry tracing being instrumented the code itself: https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/src/balancereader/pom.xml#L69

This will require extra work to 1) add OT tracing to the ledgermonolith source code, and 2) handle the env variable in the source code. 

If all you want is to get around the skaffold error-- and not actually add tracing to the ledgermonolith--  can you post more logs in the github issue so I can see what's going on? Thanks!

PR #411, opened 2020-11-27
LedgerMonolith refactor

### LINKED ISSUE #329: LedgerMonolith send payment records duplicate transactions
Action:
Deploy LedgerMonolith variant
Log into user account
Send a payment

Expected:
One payment is recorded

Actual:
Two identical payments are recorded

askmeegs:
Fixes #329 and refactors the LedgerMonolith to better function as one unified service. 

- [x] Combines the 3 controllers into one `LedgerMonolithController`, effectively concatenating the controller endpoints of the 3 microservices 
- [x] Combines the two read caches (`BalanceCache`, `TransactionHistory`) into one `LedgerReaderCache` . This seems to have resolved the duplicate transaction error #329 (screenshot below of current functionality) 

Screen Shot 2020-11-27 at 2 44 58 PM

- [x] Removed the out-of-process API call from ledgerwriter to balancereader, this is unnecessary now that they're all in the same application. Replaced this with a local call to the `LedgerReaderCache`.   
- [x] Removes the copied manifests / dedicated skaffold file. Can just use a subset of the upstream manifests with a modified config.yaml pointing at the monolith.  
- [x] Removed the firewall. When I tested pod-to-GCE connectivity using the `internal` DNS name of my instance as K8s config, **I did not need an extra firewall rule.** If this turns out to not be the case, every firewall rule I tried to create was immediately removed by the GCE enforcer, so we'd need to use the [fwupdater hack](https://github.com/GoogleCloudPlatform/bank-of-anthos-scripts/tree/b5bdef188482328b295f573c33280d1835e5091d/install/kops/firewall-updater) or similar to periodically re-create. 
- [x] rebuild + push canonical artifacts to GCS


src/ledgermonolith/src/main/java/anthos/samples/bankofanthos/ledgermonolith/LedgerReaderCache.java
```diff
@@ -0,0 +1,91 @@
...
+            
+            LOGGER.info(String.format("💡 Cache got from DB: accountId: %s, balance: %s, %s transactions",
+            accountId, balance.toString(), txns.size()));
+            return new AccountInfo(balance, txns); 
+          }
+        };
+      return CacheBuilder.newBuilder()
+          .recordStats()
+          .maximumSize(expireSize)
+          .expireAfterWrite(1, TimeUnit.SECONDS)
```
askmeegs, line 88: (this is the cache hack I was referring to, which hardcodes keys to expire in 1 second)

src/ledgermonolith/src/main/java/anthos/samples/bankofanthos/ledgermonolith/LedgerMonolithController.java
```diff
@@ -16,36 +16,176 @@
...
+                && this.ledgerReaderCache.asMap().containsKey(fromId)) {
+                AccountInfo info = ledgerReaderCache.asMap().get(fromId); 
+                Long prevBalance = info.getBalance(); 
+                Long newBalance = prevBalance - amount; 
+                processTransaction(fromId, newBalance, transaction);
+            }
+
+            if (toRouting.equals(localRoutingNum)
+                && this.ledgerReaderCache.asMap().containsKey(toId)) {
+                AccountInfo info = ledgerReaderCache.asMap().get(fromId);
```
askmeegs, line 146: @djmailhot  I found the bug in my implementation, it was getting the wrong value from the cache. This should have been `toId`, not `fromId`. :)

PR #412, opened 2020-11-30
Ledgermonolith install script

leonidv-vls:
Change summary:
Currently, the creation of ledgermonolith-service VM is
tailored for faster demo, by using startup-script.sh which
covers both installation activities and service startup.
This approach, however, has two main drawbacks as the script:
- installs / updates binaries on every VM start
- initializes / generates DB on every VM start

New version of the script:
1. Checks if service has been already installed
2. If not:
- installs pre-requisites
- downloads binaries
- initializes DB
- installs enables systemd service to run ledgermonolith app

Thanks!

### DIFF
--- a/src/ledgermonolith/init/install-script.sh
+++ b/src/ledgermonolith/init/install-script.sh
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 
-# Startup script to start the ledgermonolith service from a JAR.
+# Boot script to install and start the ledgermonolith service from a JAR.
 #
 # Expects build artifacts to be available on Google Cloud Storage.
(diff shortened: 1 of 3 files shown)


askmeegs:
Hi Leonid, unfortunately our CI doesn't support forks (which is why the build is failing). I've given you branch write access to this upstream repo - could you please put out a new PR using a direct branch? Sorry about that! 

Also a heads-up that we're refactoring the ledgermonolith source code to fix a few major bugs - #411 - so if it's alright, we may hold reviewing this until the refactor merges in. Thanks!

leonidv-vls:
Sure, will drop this one and create the branch.
Thanks!

PR #414, opened 2020-12-01
Ledgermonolith install script

leonidv-vls:
Currently, the creation of ledgermonolith-service VM is
tailored for faster demo, by using startup-script.sh which
covers both installation activities and service startup.
This approach, however, has two main drawbacks as the script:

installs / updates binaries on every VM start
initializes / generates DB on every VM start
New version of the script:

Checks if service has been already installed
If not:
- installs pre-requisites
- downloads binaries
- initializes DB
- installs enables systemd service to run ledgermonolith app

Thanks!

### DIFF
--- a/src/ledgermonolith/init/install-script.sh
+++ b/src/ledgermonolith/init/install-script.sh
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 
-# Startup script to start the ledgermonolith service from a JAR.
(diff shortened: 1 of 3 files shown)


askmeegs:
Thanks for this! I'm going to merge this into the larger Monolith Refactor PR, test it, and it will merge into master alongside #411  :)

leonidv-vls:
Great, thanks!

Issue #447, opened 2021-02-05
Prod cluster frontend readiness probe fails every 3-4 days


askmeegs:
```
Events:
  Type     Reason     Age                   From     Message
  ----     ------     ----                  ----     -------
  Warning  Unhealthy  4m28s (x61 over 26h)  kubelet  Readiness probe failed: Get http://10.44.1.40:8080/ready: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

Shabirmean:
Issue update:

Mostly seeing socket errors during and around the time when the probe fails:
```
2021-02-11 17:28:41 | [ERROR] | exception | Socket error processing request.
Traceback (most recent call last):
  File "/env/lib/python3.7/site-packages/gunicorn/workers/gthread.py", line 266, in handle
    req = next(conn.parser)
  File "/env/lib/python3.7/site-packages/gunicorn/http/parser.py", line 41, in __next__
    self.mesg = self.mesg_class(self.cfg, self.unreader, self.req_count)
  File "/env/lib/python3.7/site-packages/gunicorn/http/message.py", line 186, in __init__
    super().__init__(cfg, unreader)
  File "/env/lib/python3.7/site-packages/gunicorn/http/message.py", line 53, in __init__
    unused = self.parse(self.unreader)
  File "/env/lib/python3.7/site-packages/gunicorn/http/message.py", line 235, in parse
    self.headers = self.parse_headers(data[:idx])
  File "/env/lib/python3.7/site-packages/gunicorn/http/message.py", line 73, in parse_headers
    remote_addr = self.unreader.sock.getpeername()
OSError: [Errno 107] Transport endpoint is not connected
```

The issue can be related to the `gunicorn` server. Many people have raised the same behaviour for `Flask` and `Django` applications using `gunicorn` here: https://github.com/benoitc/gunicorn/issues/1913

Most of them reported the intermittent behaviour in `AppEngine` and `CloudRun` in GCP for their Apps. Someone actually discovered the issue (https://github.com/benoitc/gunicorn/issues/1913#issuecomment-601255354) to be a race-condition causing the server to crash once every now and then and opened a PR: https://github.com/benoitc/gunicorn/pull/2277

The PR was merged but there has not been a release of a newer version of `gunicorn` yet. They are still discussing and according to this thread (https://github.com/benoitc/gunicorn/issues/1913#issuecomment-762490975) they have asked to Watch for a release.

[...]

PR #456, opened 2021-02-12
Add liveness probe to the frontend container

bourgeoisor:
Improvement for #447 

**Change summary:**
- Add liveness probe to the frontend container, such that any non-responsiveness in the API for more than 30 seconds should result in the container getting restarted.

**Remaining issues / concerns:**
- There still exist an underlying issues (potentially with `gunicorn`) surrounding the frontend API not being responsive that should be addressed (thus, let's leave #447 opened.)
- Open question: The timeouts I have chosen for the liveness probe are fairly arbitrary in choosing (though intentionally longer than the readiness probe, since we don't want any unnecessary reboots if for example there was just a higher load than usual on the frontend and that the API is just taking a couple seconds to reply)
- Thought: The liveness probe does the same check than the readiness probe. That doesn't cause in issues per-se, it is just a bit redundant. We could think of improving the readiness probe down the line to do some more dependency-based checks.

**Testing:**
I have tested this by deploying my branch onto a GKE cluster:
```
~ kubectl get pods | grep frontend
frontend-84cd68f957-rk9n8             1/1     Running   0          7m1s
```

```
~ kubectl describe pod/frontend-84cd68f957-rk9n8
    Liveness:   http-get http://:8080/ready delay=60s timeout=30s period=15s #success=1 #failure=3
    Readiness:  http-get http://:8080/ready delay=10s timeout=10s period=5s #success=1 #failure=3
```

And then `exec` into the frontend container to find the API sub-process and kill it (`ps` and `kill -9`), which resulted in a reboot a few seconds later:
```
~ kubectl get pods | grep frontend
frontend-84cd68f957-rk9n8             0/1     Running   1          19m
```
```
~ kubectl get pods | grep frontend
frontend-84cd68f957-rk9n8             1/1     Running   1          19m
```

### DIFF
--- a/dev-kubernetes-manifests/frontend.yaml
+++ b/dev-kubernetes-manifests/frontend.yaml
@@ -74,6 +74,13 @@ spec:
           initialDelaySeconds: 10
           periodSeconds: 5
           timeoutSeconds: 10
+        livenessProbe:
+          httpGet:
+            path: /ready
+            port: 8080
+          initialDelaySeconds: 60
+          periodSeconds: 15
+          timeoutSeconds: 30
         resources:
           requests:
             cpu: 100m
--- a/kubernetes-manifests/frontend.yaml
+++ b/kubernetes-manifests/frontend.yaml
@@ -74,6 +74,13 @@ spec:
           initialDelaySeconds: 10
           periodSeconds: 5
           timeoutSeconds: 10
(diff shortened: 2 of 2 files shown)


Shabirmean (approved):
LGTM! 🚀

askmeegs (approved):

PR #457, opened 2021-02-17
Exit on error ben

benburdick:
Change summary:
- exits if there is an error applying manifests (otherwise it can fail to deploy and give you a success message anyway)
- updated opsani readme to explicitly add the directory you should be running deploy.sh from (it fails to find kubernetes-manifests dir if not run from opsani/)


benburdick:
mistake

PR #460, opened 2021-02-25
Add gh actions label

Shabirmean:
Fixes # 
Issues with creating a new host for use by GitHub Actions pipeline. Without this change when a new host is being added to the GitHub Actions pipeline - _before it is fully setup and ready to take on jobs_ - incoming events may trigger a workflow on that un-ready host. 

To evade this, it is necessary to ensure incoming events are always scheduled on existing hosts and the new one must be used only when it is fully ready.

**_Change summary:_**
- Add a **live** label to all the Github actions workflows
- Also add the same label to the existing hosts under `Settings -> Actions`

### DIFF
--- a/.github/workflows/ci-master.yaml
+++ b/.github/workflows/ci-master.yaml
@@ -21,7 +21,7 @@ on:
       - release/*
 jobs:
   code-tests:
-    runs-on: self-hosted
+    runs-on: [self-hosted, live]
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


bourgeoisor (approved):
This LGTM.
Were the prod runners tagged appropriately?

Shabirmean:
> This LGTM.
> Were the prod runners tagged appropriately?

Yup. The last one is the new host.

image

PR #463, opened 2021-03-04
Python/Flask - Replaces gunicorn with uWSGI 

askmeegs:
**Note** - I created an uptime check for my own long-standing deployment of this PR. Before merging, I want to make sure the readiness probe issues don't recur. **Given no probe failures, this PR will be safe to merge on Thursday, 3/18**  

Fixes the ongoing readiness + liveness probe issues for the Python services (#447), which seem to be caused by this [gunicorn sockets bug](https://github.com/benoitc/gunicorn/issues/1913). While the gunicorn bug has been fixed, they haven't put out a release yet, and haven't since 2019. After conferring w/ team, seems best to switch the [Flask WSGI server to uWSGI](https://flask.palletsprojects.com/en/1.1.x/deploying/wsgi-standalone/) - this doesn't affect the Flask app itself or the Python APIs. Just the server fronting the Flask app.

### Changelog 

(for each Python service - contacts, userservice, frontend)  

- [x] replaces gunicorn server startup `CMD` in Dockerfile with uwsgi server startup 
- [x] removes logging.conf - flask log date formatting now inline in python
- [x] adds uwsgi.ini which is the uwsgi config file. hardcodes # threads (same # as we used for gunicorn), date format, etc. -- port configurable from dockerfile.


src/contacts/uwsgi-conf.ini
```diff
@@ -0,0 +1,10 @@
+[uwsgi]
+uid = root
+gid = root
+
+processes = 4
+master = true
+module = contacts:create_app()
+die-on-term = true
+
+disable-logging=True
```
askmeegs, line 10: Note - this logging only refers to the uwsgi access logs, which happen on every request and were noisy. And I couldn't find a way to match them to the user-provided log level, as we do with the flask app logs. So I turned them off for now. 

But the Flask logs are intact / unchanged, still respond to log level.

src/userservice/userservice.py
```diff
@@ -217,11 +217,31 @@ def _shutdown():
...
-    app.logger.setLevel(logging.getLogger('gunicorn.error').level)
-    app.logger.info('Starting userservice.')
-
+    # set log formatting
+    date_format = "%Y-%m-%d %H:%M:%S"
+    message_format = '%(asctime)s | [%(levelname)s] | %(funcName)s | %(message)s'
+    logging.basicConfig(format= message_format, datefmt= date_format, stream=sys.stdout)
+
+    # set log level
+    log_levels = {
```
askmeegs, line 226: This way of setting the log level does seem clunky to me. Before, we let gunicorn parse the user-provided log level string, seems uwsgi doesn't support that.

Shabirmean (approved):
LGTM! 👌🏼 

Does the PR deployments also have the `loadgenerator` running?

.pylintrc
```diff
@@ -5,4 +5,5 @@ disable=F0401,  # import loading
         E1101,  # function members
         R0801,  # duplicate code
         R0903,  # public methods
-        R0915   # too many statements
+        R0915,   # too many statements
+        R0914   # too many local variables
```
bourgeoisor, line 9: Are we using an open source Google style guide for the linter?
If so, we should strongly consider disabling as few rules as possible, else it sorts of defeats the purpose (of the linter).
I'm not too sure about pylint specifically but there should also be a way to disable a certain rule _locally_, for a tiny chunk of code where the rule just doesn't make sense, while keeping it enabled elsewhere.
Thoughts?

src/contacts/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
+# Copyright 2021 Google LLC
```
bourgeoisor, line 1: General question: do we typically modify the copyright year per-file on-change, or do we try to keep all of them up-to-date as much as possible?

bourgeoisor (approved):
LGTM! Just a general question as well as a comment on the linter rules.

.pylintrc
askmeegs: Filed a follow-up issue for this! https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/465

src/contacts/Dockerfile
askmeegs: I try to manually update to the current year anytime I change the file. The `dev-kubernetes-manifests` have to get manually updated for the new year, before generating the next release

askmeegs:
> LGTM! 👌🏼
> 
> Does the PR deployments also have the `loadgenerator` running?

Yes!

PR #482, opened 2021-04-15
Add instructions for Frontend + Ingress + TLS  

### LINKED ISSUE #478: HTTP to HTTPS redirect doesn't work for /login landing page
The `frontend` currently serves requests via a [`redirect`](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/src/frontend/frontend.py#L353) _(I am not completely sure why that is; maybe a Flask thing 🤔 )_, thus regardless of the request URL (`http` or `https`), the response is a `redirect` to a URL constructed with `Flask.url_for()`. 

The url scheme used in `Flask.url_for()` is based on the environment var `SCHEME` as [seen here](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/src/frontend/frontend.py#L517)
```python
app.config['SCHEME'] = os.environ.get('SCHEME', 'http')
```

With this, the way to enable `HTTPS` is to set the environment variable `SCHEME=https` on the `frontend` pod. However still, the initial landing page `GET /login` does not use this redirect format and just renders the `index.html` as [seen here](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/src/frontend/frontend.py#L357). Thus, even with `SCHEME=https` the landing page is served over `http://` if the request is just `http://`.

Whilst this is something that could be handled at the ingress-LB level, this is not ideal to leave it as is in the app. I.E. if the scheme is explicitly set to `https` then the app should always redirect to `https`. Also this can be frictious for cases like Multi Cluster Ingress (MCI) being used whilst MCI still does not supported `http->https` redirects.

**Potential solution fetched from [here](https://stackoverflow.com/questions/45742972/forcing-use-of-https-with-flasks-before-request):**
```python
@app.before_request
def https_redirect():
    if (request.endpoint in app.view_functions and 
         os.environ.get('SCHEME') is 'https' and 
         not request.is_secure):
        return redirect(request.url.replace('http://', 'https://'))
```

askmeegs:
**Note** - does not close #478 as it doesn't solve the multicluster ingress + force HTTPS issue. 

Adds manfiests in the `extras/` folder + instructions for enabling frontend HTTPS on a global static IP, using a self signed cert, Ingress resource, and FrontendConfig.


extras/tls-ip-selfsigned/frontend-ingress.yaml
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: frontend-ingress
+  annotations:
+    kubernetes.io/ingress.global-static-ip-name: boa-single-cluster-ip
+    # kubernetes.io/ingress.allow-http: "false"
```
Shabirmean, line 20: We can remove this comment

Shabirmean (commented):
The Single Cluster setup is looking good. Let's see what's the feedback on MCI from Networking team

extras/tls-cloudsql-multicluster/multicluster-ingress-https.yaml
```diff
@@ -0,0 +1,42 @@
...
+        protocol: TCP
+        port: 8080
+        targetPort: 8080
+---
+apiVersion: networking.gke.io/v1
+kind: MultiClusterIngress
+metadata:
+  name: frontend-global-ingress
+  annotations:
+    networking.gke.io/static-ip: 34.102.238.83
```
Shabirmean, line 34: This IP should be changed to have the placeholder `STATIC_IP`

extras/tls-cloudsql-multicluster/setup.sh
```diff
@@ -0,0 +1,92 @@
...
+kubectx cluster2
+kubectl apply  -n ${NAMESPACE} -f ../cloudsql/kubernetes-manifests
+kubectl delete svc frontend
+
+echo "Creating a static IP..."
+gcloud compute addresses create boa-multi-cluster-ip --global
+IP=`gcloud compute addresses describe boa-multi-cluster-ip --global --format="value(address)"`
+
+echo "Injecting the static IP value into the MCI resource..."
+gsed -i "s/STATIC_IP/${IP}/g" multicluster-ingress-https.yaml
```
Shabirmean, line 77: I think we have to mention `gsed` in the pre-requisite. Also maybe just use `sed`.
```
sed -i -e "s/STATIC_IP/${IP}/g" multicluster-ingress-https.yaml
```

Shabirmean (approved):

PR #490, opened 2021-05-08
Split single skaffold config into modularized skaffold configs

gsquared94:
## Motivation:
Skaffold recently released two features:
- [multiple configuration support](https://skaffold.dev/docs/design/config/#multiple-configuration-support) that allows defining more than one `build-sync-deploy` pipeline in the same `skaffold.yaml` document.
- [configuration dependencies](https://skaffold.dev/docs/design/config/#configuration-dependencies) that allows defining dependencies between different skaffold configurations.

These two features together allow hierarchical build and deploy operations for multi-microservice and multi-repository applications. It also allows running individual _modules_ instead of the previous all-or-nothing approach.

Two examples of this being used are forks of this repo:
- https://github.com/gsquared94/bank-of-anthos-demo
- https://github.com/gsquared94/skaffold-remote-configs-demo

Note that we still have a root-level `skaffold.yaml` that allows us to run the following with no problems:
```
skaffold dev --default-repo=gcr.io/$PROJECT_ID/bank-of-anthos
```

## Change summary:
- Update skaffold config version to latest.
- Split into multiple skaffold modules.
- Update cloudshell tutorial to not include manual step.

Remaining issues / concerns:
- Cloudshell tutorial should elaborate on being able to trigger individual modules using the `--module` flag.


gsquared94:
cc. @viglesiasce, @peterlin741, @nkubala

gsquared94:
what's the process to update `skaffold` binaries in the test runners?

askmeegs:
Hi Gaurav and team, could you add some more details about the benefits of splitting up the skaffold config? With one skaffold config I can change on microservice's source code and iterate on it fine. 
I'm also a bit concerned about the complexity of having a lot of different skaffold configs instead of just one. 

I wonder if this could be a separate code sample instead? or be added to the `extras/` folder? 


+ the skaffold binaries in the test runners are installed via a dependencies script, we'd have to manually upgrade skaffold in our GCE custom Github actions runners. https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/.github/workflows/install-dependencies.sh#L43

Shabirmean:
> Hi Gaurav and team, could you add some more details about the benefits of splitting up the skaffold config? With one skaffold config I can change one microservice's source code and iterate on it fine.
> I'm also a bit concerned about the complexity of having a lot of different skaffold configs instead of just one.
> 
> I wonder if this could be a separate code sample instead? or be added to the `extras/` folder?
> 
> * the skaffold binaries in the test runners are installed via a dependencies script, we'd have to manually upgrade skaffold in our GCE custom Github actions runners. https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/.github/workflows/install-dependencies.sh#L43

@gsquared94 any updates on this?

gsquared94:
> @gsquared94 any updates on this?

Sorry for the late reply. I was out for the past 2 weeks.

> Hi Gaurav and team, could you add some more details about the benefits of splitting up the skaffold config? 

I elaborated on the description. The splitting up of the skaffold config mainly provides a pattern to be able to split up the `build-sync-deploy` setup for the different components and even support hosting them in their own repositories. I have linked two forks in the description that do this, however this repo being a monolith it isn't immediately helpful this way. But due to its discoverability it can be a good starting point for people looking for using this feature in a non-trivial app like this.

> With one skaffold config I can change one microservice's source code and iterate on it fine.

But it requires you to run all the services that are listed in the single `skaffold.yaml`. The refactor provides a way to be able to only run a single service via `skaffold dev --module <module_name>`.

> I'm also a bit concerned about the complexity of having a lot of different skaffold configs instead of just one.

Right now I've defined a skaffold config file per microservice. I can reduce it to only a `skaffold.frontend.yaml` and a `skaffold.backend.yaml`. 

> I wonder if this could be a separate code sample instead? or be added to the `extras/` folder?

I can move the entire thing inside `extras/`. I initially thought to replace the existing skaffold config, but if you think that this is better than let me know.

> the skaffold binaries in the test runners are installed via a dependencies script, we'd have to manually upgrade skaffold in our GCE custom Github actions runners. https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/.github/workflows/install-dependencies.sh#L43

The failed CI seems to suggest that the binary hasn't been upgraded to latest release since it couldn't recognise the schema version. Could you do the manual upgrade? The current schema version is over a year old.

askmeegs:
Thanks for these responses! I think we should still have one skaffold file as the default for the app, as running all the services is the most common path both for deployment and development (we don't provide mocked APIs so the only way to develop + test the frontend is by actually running the backends). 

I recommend forking this repo and keeping the multi module skaffold configs in the fork. (or, if this is for the sample for specific tool - either skaffold or cloud code - maybe putting the configs in the example repos for those tools.) 

We can discuss putting this in `extras/` long term, but we haven't had anyone request this functionality yet, and we try not to have too many samples in the upstream repo. Happy to link to another repo in this repo's README, though.

gsquared94:
Thanks for the clarification. Closing this for now.

viglesiasce:
Hey friends! 

Sorry for not chiming in earlier here....


> Thanks for these responses! I think we should still have one skaffold file as the default for the app, as running all the services is the most common path both for deployment and development (we don't provide mocked APIs so the only way to develop + test the frontend is by actually running the backends).

As we modularize we would still have a top level skaffold that ties together the modules to provide the same experience as today for those who want it.

Guarav has also shown the ability to slice a set of services from the DAG and use that as a way to work on only on a subset.

Since this repo is an example of many of our features/functionality in GCP for developing microservices, we want to make sure that we are exposing users to the features that we've built in Skaffold (and soon Cloud Code) for making that more productive.

gsquared94:
Closing this in favor of #514

PR #493, opened 2021-05-21
Python - adds “callable=app” to uwsgi.conf 

### LINKED ISSUE #492: "No python application" found error (Service mesh)
When running the app with Anthos Service Mesh or Istio, sometimes the Python services fail with this error -- 

```
contacts-d5dcdc87c-gmmhc contacts --- no python application found, check your startup logs for errors ---
```

Restarting the pods manually usually resolves the error. 

Not sure what exactly is causing this, possibly a config error in the UWSGI config. https://stackoverflow.com/questions/50305654/uwsgi-flask-no-python-application-found-check-your-startup-logs-for-errors

askmeegs:
Fixes #492 

#### Change summary:

Fixes a bug that seems to show up when Bank of Anthos is deployed with Istio or ASM, where on startup, the uwsgi server can’t find the flask “app.” 

```
userservice-f6996b4c8-j7njb userservice --- no python application found, check your startup logs for errors ---
```

Turns out uwsgi by default looks for “application,” not “app” (as we’re using). Seems that the simplest way to fix is to add a “[callable](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#callable)” config to the uwsgi conf which tells uwsgi to look for “app,” not “application.”  When I add this to each of the 3 python services’ config, it fixes the problem.

Source - https://code-maven.com/deploying-python-flask-using-uwsgi-on-ubuntu-14-04

#### Remaining issues / concerns:

Haven’t gotten to the bottom of why I’ve only seen this problem when the app has the Istio sidecar proxy. It might have to do with the different startup rate? Or something about shared filesystems between the proxy container and the server containers?


Shabirmean (approved):
LGTM!

PR #501, opened 2021-06-09
Use pure python entrypoints to unbreak skaffold debugging

### LINKED ISSUE #500: python services do not work with `skaffold debug`, cause status check failure.
- Since `uwsgi` is a binary and not pure python, the skaffold python debugger is unable to detect python sessions to debug. This case applies to all flask services in this sample. Running `skaffold debug` will result in `error launching python debugging: unable to determine python version from \"uwsgi\"`.
- The same reasoning applies to bash script files. When a bash script is called as the entrypoint that in turn calls python code, running `skaffold debug` will result in `error launching python debugging: unable to determine python version from \"/bin/bash\"`. This applies to the `loadgenerator` service in this sample,

j-windsor:
Fixes #500.

Change summary:
- Reverts #463 and bumps Gunicorn version to address the concern that prompted the switch to `uwsgi`.
- loadgenerator now calls `locust` directly instead of from a bash script. Dockerfile was modified to support debugging locust files.

Debugging will still not work until this is fixed on the skaffold side: https://github.com/GoogleContainerTools/container-debug-support/issues/83


askmeegs:
Hi Jay, could you please provide a bit more context on why this PR is needed?
Gunicorn gave us some issues related to readiness/liveness probes and that was what prompted the move from Gunicorn to UWSGI. 

https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/447

It does seem like the underlying bug that caused our gunicorn issues -- https://github.com/benoitc/gunicorn/issues/1913 -- has been fixed, and gunicorn has put out an updated release. but I'd rather not do another backend overhaul unless there's a concrete benefit behind doing so.

j-windsor:
Hi @askmeegs! I provided some rationale and encountered error messages in the linked issue, but the TLDR is that running `skaffold debug` or running debug in Cloud Code will not work with this sample. Skaffold currently just fails and exits (@briandealwis has a bug to allow it to fail silently instead: https://github.com/GoogleContainerTools/container-debug-support/issues/81), and it would be much more ideal if users can actually link their debugger to this sample, and if it worked out of the box with skaffold and Cloud Code.

src/userservice/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2021 Google LLC
+# Copyright 2019 Google LLC
```
kelsk, line 1: nit: license should be 2021

j-windsor: I updated all touched files to have the correct copyright year.

j-windsor:
Thanks for the review @kelsk! I addressed your comment. PTAL.

kelsk:
@j-windsor perfect, thanks!

@askmeegs For a little more context, the main goal of configuring Bank of Anthos to be debuggable using Cloud Code/`skaffold debug` is to let folks use the Cloud Code debug feature without additional setup. 

If it's not feasible to revert back to using `gunicorn` in the near future, maybe we could keep this version in a specific branch and have that branch cloned when a user opens BoA in Cloud Shell. (We could add a one-liner instruction for users to clone the specific branch if they want to try debugging on Cloud Code in their IDE.) WDYT?

askmeegs:
Hi Kelsey and Jay, after conferring with the team, we've decided to revert the python services from uwsgi back to gunicorn, since gunicorn has released a bugfix for the issue that caused us to move off it in the first place. 

Because this involves a bit more than what's in this PR (eg. re-adding a logging.conf) if it's alright I'm going to close this PR and put out a new one with all the changes. I will add you both as reviewer.

askmeegs (approved):
Tested locally in a GKE cluster, and a cluster with Anthos Service Mesh. LGTM. 

No readiness probe issues observed. (which was the gunicorn bug that caused us to move to uwsgi - ). Will need to keep an eye on this. https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/447

j-windsor:
Thanks @askmeegs and @kelsk! Feel free to merge whenever.

PR #514, opened 2021-07-20
Split single skaffold config into modularized skaffold configs

gsquared94:
This PR is a rework of #490 incorporating some of the review feedback.

## Motivation:
Update this project's use of `skaffold` to incorporate two new features:
- [multiple configuration support](https://skaffold.dev/docs/design/config/#multiple-configuration-support) that allows defining more than one `build-sync-deploy` pipeline in the same `skaffold.yaml` document.
- [configuration dependencies](https://skaffold.dev/docs/design/config/#configuration-dependencies) that allows defining dependencies between different skaffold configurations.

These two features together allow hierarchical build and deploy operations for multi-microservice and multi-repository applications. It also allows running individual _modules_ instead of the previous all-or-nothing approach.

Two examples of this being used are forks of this repo:
- https://github.com/gsquared94/bank-of-anthos-demo
- https://github.com/gsquared94/skaffold-remote-configs-demo
---
In this PR, we separate out 5 modules:
- a `setup` module to deploy the JWT secret and config maps
- a `db` module that deploys the two database services `accounts-db` and `ledger-db`
- a `backend` module that deploys all the backend services `balance-reader`,`contacts`,`ledger-writer`,`transaction-history` and `userservice`.
- a `frontend` module that deploys the `frontend` service
- a `loadgenerator` module that deploys the `loadgenerator` service

The dependency between the modules is setup using the `requires` keyword in each module definition.

---
With this change, we can still run the following with no problems:
```
skaffold dev --default-repo=gcr.io/$PROJECT_ID/bank-of-anthos
```

But now we can also run a subset of the application using the `--module` flag.
Run only the database services:
```
skaffold dev --default-repo=gcr.io/$PROJECT_ID/bank-of-anthos --module db
```
Run the app without a frontend:

```
skaffold dev --default-repo=gcr.io/$PROJECT_ID/bank-of-anthos --module backend
```

Scoping the run by modules still honors the dependency graph setup using the `requires` keyword across modules, so invoking `backend` module will also invoke `db` and `setup` modules.
 
## Change summary:
- Update skaffold config version to latest.
- Split into multiple skaffold modules.

Remaining issues / concerns:
- Cloudshell tutorial should elaborate on being able to trigger individual modules using the `--module` flag once supported in CloudCode IDEs.

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -12,12 +12,46 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-apiVersion: skaffold/v2alpha4
+apiVersion: skaffold/v2beta18
 kind: Config
+metadata:
+  name: setup # module defining setup steps
+deploy:
+  kubectl:
+    manifests: 
+    - extras/jwt/jwt-secret.yaml
+    - dev-kubernetes-manifests/config.yaml
+
+---
+
+apiVersion: skaffold/v2beta15
+kind: Config
+metadata:
+  name: db # module defining database deployments
+requires:
+  - configs: [setup]
 build:
   artifacts:
-  - image: gcr.io/bank-of-anthos/frontend
-    context: src/frontend
+  - image: gcr.io/bank-of-anthos/accounts-db
+    context: src/accounts-db
+  - image: gcr.io/bank-of-anthos/ledger-db
+    context: src/ledger-db
+deploy:
+  kubectl:
+    manifests:
+    - dev-kubernetes-manifests/accounts-db.yaml
+    - dev-kubernetes-manifests/ledger-db.yaml
+
+---
+
+apiVersion: skaffold/v2beta18
+kind: Config
+metadata:
+  name: backend # module defining backend services
+requires:
+  - configs: [db]
+build:
+  artifacts:
   - image: gcr.io/bank-of-anthos/ledgerwriter
     jib:
       project: src/ledgerwriter
@@ -31,23 +65,45 @@ build:
     context: src/contacts
   - image: gcr.io/bank-of-anthos/userservice
     context: src/userservice
-  - image: gcr.io/bank-of-anthos/loadgenerator
-    context: src/loadgenerator
-  - image: gcr.io/bank-of-anthos/accounts-db
-    context: src/accounts-db
-  - image: gcr.io/bank-of-anthos/ledger-db
-    context: src/ledger-db
-  tagPolicy:
-    gitCommit: {}
-  local:
-    concurrency: 0
 deploy:
-  statusCheckDeadlineSeconds: 300
   kubectl:
     manifests:
-    - ./dev-kubernetes-manifests/**.yaml
-    # deploy a pre-build secret
-    # in practice, this should not be checked in to source control
-    - ./extras/jwt/jwt-secret.yaml
-profiles:
-  - name: local-code
+    - dev-kubernetes-manifests/balance-reader.yaml
+    - dev-kubernetes-manifests/contacts.yaml
+    - dev-kubernetes-manifests/ledger-writer.yaml
+    - dev-kubernetes-manifests/transaction-history.yaml
+    - dev-kubernetes-manifests/userservice.yaml
+
+---
+
+apiVersion: skaffold/v2beta18
+kind: Config
+metadata:
+  name: frontend # module defining frontend service
+requires:
+  - configs: [backend]
+build:
+  artifacts:
+  - image: gcr.io/bank-of-anthos/frontend
+    context: src/frontend
+deploy:
+  kubectl:
+    manifests:
+    - dev-kubernetes-manifests/frontend.yaml
+
+---
+
+apiVersion: skaffold/v2beta18
+kind: Config
+metadata:
+  name: loadgenerator # module defining a load generator service
+requires:
+  - configs: [backend]
+build:
+  artifacts:
+  - image: gcr.io/bank-of-anthos/loadgenerator
+    context: src/loadgenerator
+deploy:
+  kubectl:
+    manifests:
+    - dev-kubernetes-manifests/loadgenerator.yaml
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -24,7 +24,7 @@ deploy:
 
 ---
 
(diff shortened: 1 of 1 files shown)


gsquared94:
This PR is a rework of https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/490 incorporating some of the review feedback.

NimJay:
As mentioned in [@gsquared94's comment](https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/490#issuecomment-837327189) from #490, we would need to upgrade the version of `skaffold` running on this repository's [GitHub Action runners](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/d1848ad56e5579703bfd425e84808ba324109814/.github/workflows/README.md#infrastructure) (virtual machines). 

@gsquared94 Do you know the best way to upgrade `skaffold`? [This script](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/d1848ad56e5579703bfd425e84808ba324109814/.github/workflows/install-dependencies.sh#L42) was likely used to initially install `skaffold` on the virtual machines.

@donmccasland Do you know what the "[central Bank of Anthos GCP project](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/d1848ad56e5579703bfd425e84808ba324109814/.github/workflows/README.md#appendix---creating-a-new-actions-runner)" is?

skaffold.yaml
```diff
@@ -12,12 +12,46 @@
...
+  name: setup # module defining setup steps
+deploy:
+  kubectl:
+    manifests: 
+    - extras/jwt/jwt-secret.yaml
+    - dev-kubernetes-manifests/config.yaml
+
+---
+
+apiVersion: skaffold/v2beta15
```
NimJay, line 27: Is there a reason we use `skaffold/v2beta15` instead of `skaffold/v2beta18`?

gsquared94:
> @gsquared94 Do you know the best way to upgrade `skaffold`? [This script](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/d1848ad56e5579703bfd425e84808ba324109814/.github/workflows/install-dependencies.sh#L42) was likely used to initially install `skaffold` on the virtual machines (Linux).

You should be able to just manually run this section of the install script again.
https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/d1848ad56e5579703bfd425e84808ba324109814/.github/workflows/install-dependencies.sh#L42-L45

skaffold.yaml
```diff
@@ -31,23 +65,45 @@ build:
...
+  - configs: [backend]
+build:
+  artifacts:
   - image: gcr.io/bank-of-anthos/loadgenerator
     context: src/loadgenerator
-  - image: gcr.io/bank-of-anthos/accounts-db
-    context: src/accounts-db
-  - image: gcr.io/bank-of-anthos/ledger-db
-    context: src/ledger-db
-  tagPolicy:
```
donmccasland, line 40: It looks like in this modular version we have lost the settings for tagPolicy, local concurrency, and deploy statusCheckDeadlineSeconds

skaffold.yaml
gsquared94: fixed

skaffold.yaml
gsquared94: the `tagPolicy` defaults to `gitCommit` so it doesn't need to be explicitly set in the configs. Same with `statusCheckDeadlineSeconds` which defaults to `600 seconds` in newer versions of skaffold (vs. the `120 seconds` it was previously).
Build concurrency also has a flag now called `build-concurrency` so the project default  can be left to build consecutively (less resource intensive default), which can be made to build in parallel by running with `--build-concurrency=0` flag.

WDYT?

NimJay:
I just upgraded `skaffold` from **1.20.0** to **1.28.0** on both GitHub Action runners (virtual machines `gh-actions-ci` and `gh-actions-ci-2`). I simply downloaded the latest `skaffold` binary and replaced the existing binary:
```
curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/latest/skaffold-linux-amd64
chmod +x skaffold
sudo mv skaffold /usr/local/bin 
```

I've re-run the checks from this PR, and it looks like they're all passing now. :)

skaffold.yaml
NimJay: I like the idea of relying on the default values (for `tagPolicy` and `statusCheckDeadlineSeconds`).
It will result in a less clustered `skaffold.yaml`

Currently looking into why `name: local-code` was introduced...

NimJay (approved):
I am going to approve this pull-request.
@donmccasland also approved it yesterday (verbally) during our weekly Anthos/GKE DPE meeting.

My only concern is the fact that anyone using Bank of Anthos will now have to upgrade their version of `skaffold`. That is, they will need at least version 1.27.0. But this required `skaffold` upgrade should be fine because the error message produced (when using an outdated version) is descriptive (and suggests upgrading `skaffold`):
```
WARN[0000] Your Skaffold version might be too old. Download the latest version (1.28.1) from:
  https://github.com/GoogleContainerTools/skaffold/releases/tag/v1.28.1
parsing skaffold config: error parsing skaffold configuration file: unknown skaffold config API version "skaffold/v2beta18". Set the config 'apiVersion' to a known value. Check https://skaffold.dev/docs/references/yaml/ for the list of valid API versions. Otherwise, check that your skaffold version is up-to-date.
```

I tested `skaffold dev` with 5 versions of skaffold:
|`skaffold` Version|||
|-|-|-|
|1.24.1|❌|Got error in snippet above.|
|1.26.0|❌|Got error in snippet above.|
|1.26.0|❌|Got error in snippet above.|
|1.27.0|✅ |Worked.|
|1.28.1|✅ |Worked.|

NimJay:
@askmeegs welcome back (from your vacation)!
Sorry, this isn't the warmest welcome, but I just want your final word/review on this change before I merge it in.
The reason I ask for your review: you had suggested alternative solutions in https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/490#issuecomment-864572609.

askmeegs (approved):

askmeegs:
Tested locally with Skaffold 1.28. LGTM.

PR #516, opened 2021-07-29
Structure logs to display correctly in Cloud Operations

### LINKED ISSUE #498: Add Cloud Logging structured logs for Java and Python services
Only seeing this mislabeled severity with `INFO` logs --> mistakenly logged as `ERROR`. 

Services where I see this: **userservice**, **contacts**, and **frontend** (all the python services), the two in-cluster databases (**accounts** and **ledger**). Java services are unaffected. 

Example: 

Screen Shot 2021-06-03 at 1 41 38 PM

Strangely, [the line of code there is `app.logger.info`](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/src/frontend/frontend.py#L394) but it's getting mislabeled in severity, somewhere.

bourgeoisor:
Fixes #498

This PR makes sure that logs are displayed with the intended severity level in Cloud Operations (i.e. avoiding false classification of logs like showing all logs as error or all logs as info regardless of severity).

## Overview

There are multiple ways to direct logs to Cloud Operations. For example, there are Google APIs and libraries that can be used specifically for this purpose. Alternatively, Operations is also able to read logs from known streams such as `stdout` and `stderr`. If read from those streams, it will classify all logs as `INFO` or all logs as `ERROR`, respectably. To avoid this, we can use structured logging (typically in JSON format) with specific keywords for fields that Operations recognizes.

These fields are documented here: https://cloud.google.com/logging/docs/structured-logging

Among others are `message` which is the summary that gets displayed in the query results, as well as `severity` which sets the level of the log within Logs Explorer.

## Open questions

**Postgres services:** By default, Postgres services logs are all directed to `stderr` (https://www.postgresql.org/docs/12/runtime-config-logging.html) which may create false negative (i.e. info-level logs displayed as errors). An alternative would be to redirect all logs to `stdout` or similar which would have the opposite problem (i.e. error-level log displayed as info). CSV is also supported, but doesn't seem to play well with Cloud Operations. I prefer sticking to `stderr` (the current default) than moving to `stdout`. There may be a way to do structured JSON logs but it doesn't seem to be natively supported by Postgres. Thoughts?

**Function names:** I have kept the function names in the summary of the logs for easily parsing it in the list of summaries, but perhaps it makes more sense to be hidden / tucked away into its own JSON field. Should we leave it in the summary or hide it in the JSON payload?

#### INFO

image

#### ERROR

image

#### INFO

image

#### ERROR

image

### DIFF
--- a/src/userservice/logging.conf
+++ b/src/userservice/logging.conf
@@ -28,6 +28,6 @@ formatter=generic
 args=(sys.stdout, )
 
 [formatter_generic]
-format=%(asctime)s | [%(levelname)s] | %(funcName)s | %(message)s
+format={"timestamp": "%(asctime)s", "message": "%(funcName)s | %(message)s", "severity": "%(levelname)s"}
 datefmt=%Y-%m-%d %H:%M:%S
 class=logging.Formatter
--- a/src/balancereader/src/main/resources/log4j2.xml
+++ b/src/balancereader/src/main/resources/log4j2.xml
@@ -19,7 +19,7 @@
     <Appenders>
         <Console name="console" target="SYSTEM_OUT">
             <PatternLayout
-                pattern="%d{yyyy-MM-dd HH:mm:ss} | [%level] | %M | %msg%n" />
+                pattern='{"timestamp": "%d{yyyy-MM-dd HH:mm:ss}", "message": "%M | %msg", "severity": "%level"}%n' />
         </Console>
     </Appenders>
     <Loggers>
--- a/src/contacts/logging.conf
+++ b/src/contacts/logging.conf
@@ -28,6 +28,6 @@ formatter=generic
 args=(sys.stdout, )
 
 [formatter_generic]
-format=%(asctime)s | [%(levelname)s] | %(funcName)s | %(message)s
+format={"timestamp": "%(asctime)s", "message": "%(funcName)s | %(message)s", "severity": "%(levelname)s"}
 datefmt=%Y-%m-%d %H:%M:%S
 class=logging.Formatter
--- a/src/frontend/logging.conf
+++ b/src/frontend/logging.conf
@@ -28,6 +28,6 @@ formatter=generic
 args=(sys.stdout, )
 
 [formatter_generic]
-format=%(asctime)s | [%(levelname)s] | %(funcName)s | %(message)s
+format={"timestamp": "%(asctime)s", "message": "%(funcName)s | %(message)s", "severity": "%(levelname)s"}
 datefmt=%Y-%m-%d %H:%M:%S
 class=logging.Formatter
--- a/src/ledgerwriter/src/main/resources/log4j2.xml
+++ b/src/ledgerwriter/src/main/resources/log4j2.xml
@@ -19,7 +19,7 @@
     <Appenders>
         <Console name="console" target="SYSTEM_OUT">
             <PatternLayout
-                pattern="%d{yyyy-MM-dd HH:mm:ss} | [%level] | %M | %msg%n" />
+                pattern='{"timestamp": "%d{yyyy-MM-dd HH:mm:ss}", "message": "%M | %msg", "severity": "%level"}%n' />
         </Console>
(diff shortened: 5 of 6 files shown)


Shabirmean:
This is awesome @bourgeoisor. Thank you for this change 🎉 

_Re: open questions:_

**Function names:** I like having the function name in the summary; so we can quickly get a sense of where the log is being spit out from. However, a thought I have is whether it will be useful to **also have** the information as a JSON field. This is mainly because, some of us might directly go into the JSON forgetting the summary and it might be useful to explicitly point out what the `function` is. Also for those who doesn't know the log format, it might not be self evident that the first string is the function name in the summary. 

**Postgres services:** This is interesting and frustrating. I believe this is okay given that this is a sample application. However, an ideal case would be to have our own log processor for the stack _(instead of using the stderr, stdout)_ streams  because we are dealing with both - apps written by us and by 3rd party _(i.e: postgres)_. But I think it is out of scope here.

NimJay (approved):
First off, dang! Thanks for that thorough pull-request description!

Regarding your open questions: 

**1. Postgres services**
> I prefer sticking to `stderr` (the current default) than moving to stdout.

I agree — I'd rather have non-error logs in `stderr` than error logs in `stdout`.

**2. Function names**

> I have kept the function names in the summary of the logs for easily parsing...

The function name prefix is definitely easy on the eye (and convenient). I like it! :)

```... "message": "%(funcName)s | %(message)s" ...```

It sounds like Shabir's suggesting an additional JSON field for the function (to be explicit). I'm happy either way.

File: src/userservice/logging.conf [hunk 1]
```diff
@@ -28,6 +28,6 @@ formatter=generic
 args=(sys.stdout, )
 
 [formatter_generic]
-format=%(asctime)s | [%(levelname)s] | %(funcName)s | %(message)s
+format={"timestamp": "%(asctime)s", "message": "%(funcName)s | %(message)s", "severity": "%(levelname)s"}
```
NimJay, line 31: Quick question about the message: could it possibly break the JSON (i.e., do we have to worry about escaping "s in the message)?

bourgeoisor: Interesting question. It doesn't break the entire logging per se, but it will break _those_ specific lines with quotes or new-line for example.

image

bourgeoisor: I'm still messing with it to see if I can avoid this problem.

bourgeoisor: I tried digging around some for e.g. Python and this is possible (escaping / removing certain characters like new-lines or quotes, for example) but it requires creating a custom handler typically.

Currently, none of the logs seem to use quotation or new-lines, so this problem doesn't arise.

With that in mind as well as thinking about the purpose of Bank of Anthos, does it make sense to add those lines of code which may not be related to Bank of Anthos and its purpose and add noise into the code? (This is more of an open question than a rhetoric, I don't know this answer to this.)

Logging docs + questions around the same topic:
https://docs.python.org/3/library/logging.html
https://stackoverflow.com/a/6692653
https://stackoverflow.com/a/13373364

bourgeoisor:
Thanks for the comments and feedback! As for the function name, it _does_ also appear in the JSON payload as well as the summary.

src/userservice/logging.conf
Shabirmean: Does the issue arise if the message has single quotes as well? Or is it only a problem is it's a double quote?

Shabirmean (approved):
LGTM!

src/userservice/logging.conf
bourgeoisor: Double quote only. Single quotes are fine!
And any of those hypothetical "log breaking" logs wouldn't break the logs inside the Pod, nor does it break any _other_ log entries in Cloud Logging. Rather, it'll just break the _one_ log that has a special character (either `\n` or `"`, because those break the JSON format for that _one_ log entry).

Issue #507, opened 2021-06-23
Cloud SQL instance is incorrectly suffixed "-multi"


NimJay:
Just reporting an extremely minor naming error that @askmeegs pointed out...

https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/368549b0edb4374765132fe21a87b91bcc37c295/extras/cloudsql/create_cloudsql_instance.sh#L16

bourgeoisor:
This has been fixed in #523. Closing.

PR #523, opened 2021-08-27
CloudSQL updates

arueth:
Fixes #
https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/507

Change summary:
- Brought /extras/cloudsql/kubernetes-manifests up to date with /kubernetes-manifests
- Updated gce-proxy from 1.19.0-alpine to 1.24.0-alpine

Remaining issues / concerns:

### DIFF
--- a/extras/cloudsql/README.md
+++ b/extras/cloudsql/README.md
@@ -50,7 +50,7 @@ gcloud container clusters create ${CLUSTER} \
 6. **Create a Cloud SQL admin demo secret** in your GKE cluster. This gives your in-cluster Cloud SQL client a username and password to access Cloud SQL. (Note that admin/admin credentials are for demo use only and should never be used in a production environment.)
 
 ```
-INSTANCE_NAME='bank-of-anthos-db'
+INSTANCE_NAME='bank-of-anthos-db-multi'
 INSTANCE_CONNECTION_NAME=$(gcloud sql instances describe $INSTANCE_NAME --format='value(connectionName)')
 
 kubectl create secret -n ${NAMESPACE} generic cloud-sql-admin \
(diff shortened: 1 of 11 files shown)


extras/cloudsql/README.md
```diff
@@ -50,7 +50,7 @@ gcloud container clusters create ${CLUSTER} \
 6. **Create a Cloud SQL admin demo secret** in your GKE cluster. This gives your in-cluster Cloud SQL client a username and password to access Cloud SQL. (Note that admin/admin credentials are for demo use only and should never be used in a production environment.)
 
 ```
-INSTANCE_NAME='bank-of-anthos-db'
+INSTANCE_NAME='bank-of-anthos-db-multi'
```
bourgeoisor, line 53: If I'm reading #507 correctly, I believe it's https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/368549b0edb4374765132fe21a87b91bcc37c295/extras/cloudsql/create_cloudsql_instance.sh#L16 that needs to change (to remove the `-multi`) rather than the other way around.
@NimJay ?

arueth: It is already `-multi` in `extras/cloudsql-multicluster` and `extras/cloudsql/create_cloudsql_instance.sh` so I thought I would just fix the one instance.

NimJay: @bourgeoisor That's correct — we wanted to remove `-multi` from `extras/cloudsql/`.

@arueth Ah! Gotcha. I didn't realize that `extras/cloudsql/create_cloudsql_instance.sh` is being used by `extras/cloudsql-multicluster`. I, personally, prefer removing the `-multi` suffix from the Cloud SQL instance names. But I don't feel too strongly about it. So your call. :)

cloud-pharaoh: @arueth I would just remove the multi from extras/cloudsql-multicluster and leave this as is. 
This script is used for a single cluster install as well so the name should not include it.

cloud-pharaoh (changes requested):
LGTM except for minor change requested for `multi`.

arueth:
Removed `-multi` as per review

bourgeoisor (approved):
LGTM!

cloud-pharaoh (approved):

PR #525, opened 2021-08-31
Update ledgermonolith service to simple

### LINKED ISSUE #524: Boot script for the ledgermonolith VM hangs
I noticed that the boot script for the [ledgermonolith](https://github.com/GoogleCloudPlatform/bank-of-anthos/tree/master/src/ledgermonolith) VM hangs on `systemctl start ...`:

image

This is due to the `ledgermonolith` systemd service which is currently configured to be a `oneshot` (see [here](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/src/ledgermonolith/init/install-script.sh#L133)).

The [systemd docs](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Options) describes the differences between the service types and notes that for `oneshot` services, "the service manager will [only] consider the unit up after the main process exits." This makes systemd hang until the process exits, which is never does (since `ledgermonolith` is supposed to act like a daemon.

We should change this process to be of type `simple` instead so that the service manager runs it in the background and does not continuously wait for it.

bourgeoisor:
Fixes #524 

Change summary:
- Update the `ledgermonolith` systemd service to be of type `simple` instead of `oneshot`

Tested and the VM still loads correctly (see transactions showing up):
image

### DIFF
--- a/src/ledgermonolith/init/install-script.sh
+++ b/src/ledgermonolith/init/install-script.sh
@@ -130,7 +130,7 @@ chmod +x ${MONOLITH_DIR}/ledgermonolith-service.sh
 
 cat <<EOF >/etc/systemd/system/${MONOLITH_SERVICE}
 [Service]
-Type=oneshot
+Type=simple
 RemainAfterExit=yes
 ExecStart=${MONOLITH_DIR}/ledgermonolith-service.sh
 
@@ -143,4 +143,3 @@ systemctl start ${MONOLITH_SERVICE}
 
 echo "Install Complete"
 exit
-


NimJay (approved):
First, thanks for summarizing the problem in https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/524 (and the relevant links)!

It seems like `simple` is the default `Type` when `ExecStart` is specified, but I'm glad that we're explicitly specifying it. 👍 

The only thing I worry about is the fact that "[the service manager will consider the unit started **immediately** after the main service process has been **forked off**](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Options)" — which I assume is some milliseconds/seconds before `java -jar ${MONOLITH_DIR}/${APP_JAR} > ${MONOLITH_LOG}` runs and stabilizes. But I doubt this will be an issue.

I am approving this pull-request because I trust you've done the appropriate testing (i.e., seen that the VM no longer hangs at `systemctl start ...`) and I believe `simple` is better than `oneshot` for this scenario.

bourgeoisor:
> The only thing I worry about is the fact that "[the service manager will consider the unit started **immediately** after the main service process has been **forked off**](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Options)" — which I assume is some milliseconds/seconds before `java -jar ${MONOLITH_DIR}/${APP_JAR} > ${MONOLITH_LOG}` runs and stabilizes. But I doubt this will be an issue.

That is correct! Since there is nothing in the init script nor any processes depending on `ledgermonolith` though, the end-result stays consistent. Ideally, we would use something like `forking` and have the `ledgermonolith` call `fork()` whenever appropriate, but this is outside the scope of this quick-fix.

PR #526, opened 2021-09-01
Fix CVE-2021-33503 and CVE-2020-8908

bourgeoisor:
This PR fixes:

**CVE-2021-33503 (high severity)** by bumping `urllib` to `>=1.26.5`
**CVE-2020-8908 (low severity)** by bumping `guava` to `>=30.1.1-jre`


src/balancereader/pom.xml
```diff
@@ -114,7 +114,7 @@
         <dependency>
           <groupId>com.google.guava</groupId>
           <artifactId>guava</artifactId>
-          <version>29.0-jre</version>
+          <version>30.0-jre</version>
```
arueth, line 117: Any reason for going with 30.0 instead of 30.1.1?

bourgeoisor: Nope! That was simply the latest major release. I've just updated it to get the `.1` minor release.

NimJay (approved):
Similar to Aaron's question, I'm just curious: is there a reason we're upgrading to `urllib` to `1.26.5` instead of `1.26.6` (which is the [latest](https://pypi.org/project/urllib3/#changes)). Either way, I'm happy with this PR.

Approved.

PR #545, opened 2021-12-10
Test PR - new CI

askmeegs:

PR #548, opened 2021-12-11
fix: catch decimal error and return fail

### LINKED ISSUE #182: UI Test: cannot contain more than 2 decimal digits

Shabirmean:
### Background 
- Whenever an invalid decimal number is input the UI does not allow the payment to be submitted. However, if someone directly calls the API then the server just fails with an exception
- They primarily fail because the invalid decimal number raises an exception which is not caught resulting in no response from server

### Additional Notes
- After investigating further, we notice that the real reason [for the specific cypress test](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/.github/workflows/ui-tests/cypress/integration/transfer_spec.js#L194-L202) to fail _(in the current codebase)_ is because that the [input field for the amount](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/src/frontend/templates/index.html#L331) has been set to **number=type**. This means there is no way a user can input an invalid number into that field because the browser will not allow it. The same happens when cypress sets an invalid value to this UI element, it automatically rounds to the closest valid decimal. Thus the transaction goes through without an issue - resulting in the test to fail. 
- So this PR removes the test completely - because that doesn't work anymore; however since someone could still submit an invalid number by calling the API directly - we catch this exception and return an error. 
- Some of the CI files have been update to match the test pr #545, in order to see if the cypress tests pass
- These changes must be reverted before merge

### Change Summary
- Catch the exception raised by the invalid number input
- Return an error message
- Remove the Cypress test that checks the UI element

### Testing Procedure
- Cypress tests should pass + not have the pending test for [the removed test](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/.github/workflows/ui-tests/cypress/integration/transfer_spec.js#L194-L202).
- You can try sending a valid request to the server, then copying the request as `Curl` from the Chrome inspect window
- Resubmit the curl with an invalid value from your terminal

### Related PRs or Issues 
- N/A


.github/workflows/ci-pr.yaml
```diff
@@ -19,7 +19,7 @@ on:
       - master
 jobs:
   code-tests:
-    runs-on: [self-hosted, live-v2]
+    runs-on: [self-hosted, runner-1]
```
Shabirmean, line 22: revert before merge

.pylintrc
```diff
@@ -5,4 +5,8 @@ disable=F0401,  # import loading
         E1101,  # function members
         R0801,  # duplicate code
         R0903,  # public methods
-        R0915   # too many statements
+        R0915,   # too many statements
```
Shabirmean, line 8: revert all changes to this file before merge

ahrarmonsur:
I manually tested using cURL commands from the terminal. Here are my observations:

Note: One note missing from your testing instructions was to increment the UUID with every request, as the submission will be rejected for non-unique transaction UUIDs.

## 1. Clean input `5.43`
I received the expected page refresh response on successful deposit:
```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>Redirecting...</title>
<h1>Redirecting...</h1>
<p>You should be redirected automatically to target URL: <a href="http://35.232.201.148/home?msg=Deposit+successful">http://35.232.201.148/home?msg=Deposit+successful</a>.  If not click the link.ahrar-
```

## 2. Inappropriate number containing 2 decimal points, `5.45.87`
I received the expected `500` error response, which was the scope of this PR
```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>500 Internal Server Error</title>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>
```

## 3. Potentially problematic: Truncating precision on more than 2 decimal digits, `5.4587`
I get a successful response, however when refreshing the app, I see that the posted transaction amount was `5.45`, as in the extra precision was truncated. I would have expected the posted amount to be rounded to `5.46`. What is your expectation here, @Shabirmean?

image

ahrarmonsur (changes requested):
The API now responds with `500` errors as expected.
One potential bug is that the API truncates decimals beyond 2 digits, rather than rounding. Depending on how the app was designed, the handler function should account for this.

The corresponding transaction through the UI is not possible, as the input validator does not allow more than 2 decimal digits.

NimJay:
Thanks for fixing this, Shabir! And thanks for the rigorous review, Ahrar! :)
Just adding my thoughts: 

1. I am wondering if should just allow-list `amount` values that match [this regex](https://stackoverflow.com/a/308124) — instead of allowing/rounding numbers such as `5.4587`. 🤔 

2. Ideally, the HTTP response for invalid `amount` values [would be `422 Unprocessable Entity`](https://stackoverflow.com/questions/7939137/right-http-status-code-to-wrong-input). But I think this is outside the scope of this pull-request.

Shabirmean:
@ahrarmonsur Thank you very much for the very detail review. I feel that that the rounding error is a bit beyond the scope of this PR which attempts to address the issue raised in #182. It feels more a like *improvement/feature-request* that needs to be handled separately.

@NimJay I think almost all the responses in the application (via the frontend) are `redirect 302` responses with specific `Failure` and `Success` messages. The redirects return you to the `home` page with the message which is rendered. So in the above case too, for an invalid request the response is a `302` with a specific failure message. I am not sure why this response via redirect strategy was used when the app was developed. But I agree if we were to surface proper error coded responses `422`  might be the best choice.

Let me know what you'll think and if we can merge this PR!

ahrarmonsur (approved):
I agree that the rounding behaviour is not related to the scope of this Issue. I spoke with Shabir, and he will create a separate issue to capture this.

The PR otherwise LGTM!

Shabirmean:
> I agree that the rounding behaviour is not related to the scope of this Issue. I spoke with Shabir, and he will create a separate issue to capture this.
> 
> The PR otherwise LGTM!

Created a new issues at: https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/590

PR #588, opened 2022-01-07
Fix release refs path to bank-of-anthos-ci

bourgeoisor:
This PR fixes the release refs path (from the old `bank-of-anthos` GCR to the new `bank-of-anthos-ci` one).

### DIFF
--- a/.github/snippet-bot.yml
+++ b/.github/snippet-bot.yml
@@ -1,0 +1,13 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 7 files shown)


NimJay (approved):
I did a quick scan for instances of `bank-of-anthos` in the repository.
It looks like this pull-request covered all instances in need of replacement.

Looks good to me. Approved.

Side note: We'll also need to replace instances of "`bank-of-anthos-master`" inside `releade/README.md`.

NimJay (approved):
Thanks, @bourgeoisor, for replacing `bank-of-anthos-master` with `bank-of-anthos-release`.

Approved yet again! :)

I realized that the `README.md` file mentions the use of a workload identity called `boa-ksa-master`. 
IDK how accurate this info is. It might be worth looking into later (e.g., correcting if needed).
Regardless, I think fixing it is outside the scope of this pull-request.
Feel free to merge this pull-request.

bourgeoisor:
It's not, but we should bring that back for the next release we do! I'll leave it in there for now--

PR #593, opened 2022-01-14
Specify JDK requirement to OpenJDK

bourgeoisor:
This PR specifies the JDK requirement to be OpenJDK (rather than Oracle JDK) due to more permissive licensing.


ckim328 (approved):
LGTM!

PR #594, opened 2022-01-16
Bump minor python dependencies suggested by renovate 

bourgeoisor:
This PR does a first pass of bumping deps suggested by https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/592

Note that I intentionally skipped hand-picked deps (especially major version bumps) in this pass since it was just too many deps changed at the same time. For major version bumps I'll take it a bit slower since code changes may be required.

I have deployed this branch on a cluster and navigated the various UI journeys to make sure everything looked good.


ckim328:
I didnt look into it too much, but it looks like the actions for python are failing, some trouble with installing some packages?

bourgeoisor:
Yep it's the test packages, which is why the deployment worked on my end-- Looking into it!

NimJay (commented):
* Thanks for doing this, @bourgeoisor.
* File changes look good to me. 👍 
  * Everything being updated is listed in #592. 
* I see that #592 list some dependencies (e.g., `google-cloud-core`, `packaging`, `pytest-cov`) more than once for different version. I'm not sure why.

* I will approve it as soon as we figure out the failing checks and I can smoke test the staging URL.
From my short research, it seems like we [may need to install wheel](https://stackoverflow.com/questions/34819221/why-is-python-setup-py-saying-invalid-command-bdist-wheel-on-travis-ci) in our Action Runners. I'm not sure if we'd have to modify [`.github/workflows/install-dependencies.sh`](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/release/v0.5.3/.github/workflows/install-dependencies.sh). If we do, we can do it this PR or another.

bourgeoisor:
- About multiple versions from Renovate: sometimes there are two versions: a minor bump, and a major bump. Renovate highly recommends the minor bump since that usually doesn't require code changes, but depending on case-by-case basis it may be tricky or impractical to do the major bump without some source code changes.

- Wheel is there. The issue was that it was trying to compile all packages from scratch rather than use the wheels for them. After some investigation and a bit of trial and error, I realized that the virtual env used for testing wasn't using the same `pip` as installed in the runners (which they themselves are up-to-date). I added a line to make sure pip is always running latest stable. This is a no-op if it is (so it's not adding any more time complexity in the workflows, unless pip is out-of-date, in which case we _should_ upgrade it anyway)

src/contacts/requirements.txt
```diff
@@ -1,107 +1,101 @@
 #
-# This file is autogenerated by pip-compile with python 3.6
+# This file is autogenerated by pip-compile with python 3.8
 # To update, run:
 #
-#    pip-compile --output-file=requirements.txt requirements.in
+#    pip-compile --output-file=/Users/obourgeois/workspace/repos/bank-of-anthos/src/contacts/requirements.txt /Users/obourgeois/workspace/repos/bank-of-anthos/src/contacts/requirements.in
```
Shabirmean, line 5: Should we redact file paths from local environment?

src/frontend/requirements.txt
```diff
@@ -1,11 +1,9 @@
 #
-# This file is autogenerated by pip-compile with python 3.6
+# This file is autogenerated by pip-compile with python 3.8
 # To update, run:
 #
-#    pip-compile --output-file=requirements.txt requirements.in
+#    pip-compile --output-file=/Users/obourgeois/workspace/repos/bank-of-anthos/src/frontend/requirements.txt /Users/obourgeois/workspace/repos/bank-of-anthos/src/frontend/requirements.in
```
Shabirmean, line 5: Same comment as above

Shabirmean (changes requested):
LGTM! Left a minor cleanup related comment

src/contacts/requirements.txt
bourgeoisor: Oops! Fixed :)

NimJay (approved):
The new `pip install --upgrade pip` lines look good to me.
Approved.
Thanks for tackling this, Olivier.

Issue #520, opened 2021-08-11
Possible memory leak in loadgenerator


askmeegs:
<img width="1793" alt="Screen Shot 2021-08-11 at 4 13 06 PM" src="https://user-images.githubusercontent.com/3137106/129096737-49a4a045-66ca-491f-8c30-fa668df8816e.png">

bourgeoisor:
Updating this issue since it's a few months old now. I can still replicate this so there's likely still a memory leak in the loadgen. Notice the brand new instance slowly creeping up from the get-go:

![image](https://user-images.githubusercontent.com/3271352/144518121-ef656111-446c-4e6e-910e-477e3567f91b.png)

bourgeoisor:
I've attempted two different solutions for this without much success.

a) I updated `locust` (the only direct dependency) from a pre-release version to latest. This had no visible effect:

<img width="1533" alt="Screen Shot 2022-02-09 at 11 28 05 AM" src="https://user-images.githubusercontent.com/3271352/153250373-c82c7e42-a40c-4941-8484-fbde46ab251e.png">

b) I also attempted to modify the Dockerfile to use the latest stable Python (1.10) and a cleaner base image:
```
FROM python:3.10

WORKDIR /usr/src/app

ENV PYTHONUNBUFFERED=0
ENV GEVENT_SUPPORT=True
ENV LOG_LEVEL info

COPY requirements.txt ./
RUN pip install --no-cache-dir -r requirements.txt

COPY locustfile.py .

ENTRYPOINT locust --host="http://${FRONTEND_ADDR}" --loglevel $LOG_LEVEL --no-web -c "${USERS:-10}" 2>&1
```

This also had no visible effect:

![image](https://user-images.githubusercontent.com/3271352/153250533-af55a7f4-32a0-4740-b8c9-cd3daf03f5be.png)

bourgeoisor:
Bumping down to p3 since the OOM happens over the course of multiple weeks, and it reboots / starts doing its loadgen immediately after like nothing happened.

bourgeoisor:
This is still an ongoing leak (though minimize through previous deps updates). Commenting to reset SLO.

mathieu-benoit:
Out of curiosity @bourgeoisor, do you know if we have the same behavior with OnlineBoutique's `loadgenerator`?

bourgeoisor:
I believe so. This is a 30-day chart of the past month:
[...]

PR #612, opened 2022-02-09
Bump locust deps

bourgeoisor:
This was an attempt at fixing a memory leak issue with the load generator (#520). Unfortunately it didn't help, but this version bump is necessary nonetheless due to #599.

I have tested this on a live cluster-- The load generator works as intended (bar the memory leak of course).

More details in #520.


NimJay (approved):
I did a quick smoke test of the staging URL (e.g., deposited/transferred funds).
Looks good to me!
Approved.

PR #619, opened 2022-02-11
Delete checkout step from cleanup workflows

bourgeoisor:
This PR removes the checkout step from the cleanup workflows. None of the cleanup steps require checking in, and attempting to do so yields errors if the branch was deleted before the cleanup job started running (race condition): https://github.com/GoogleCloudPlatform/bank-of-anthos/actions/workflows/cleanup.yaml

I will test this by deleting the branch immediately following merging.

### DIFF
--- a/.github/workflows/cleanup.yaml
+++ b/.github/workflows/cleanup.yaml
@@ -25,7 +25,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


ckim328 (approved):

bourgeoisor:
Verified that this worked: https://github.com/GoogleCloudPlatform/bank-of-anthos/runs/5161984029?check_suite_focus=true

PR #634, opened 2022-02-15
Add Cloud Armor in front of this deployed demo

mathieu-benoit:
Document the way to setup Cloud Armor in front of the public endpoint of this demo [bank-of-anthos.xyz](https://bank-of-anthos.xyz).

_Note: not yet actually deployed, opening this PR for review first._

What's in this PR:
- Porting the [Cloud Armor setup we now have in Online Boutique](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main/.github/release-cluster) for consistency between the 2 repos
- Document the steps to deploy the `prod-cluster` part with more automation with `gcloud`
- Add `BackendConfig` to bind Cloud Armor to the `Ingress`
- Add `FrontendConfig` to bind the SSL Policy and redirect `http` to `https`
- Update `Ingress`'s `apiVersion` from `apiVersion: networking.k8s.io/v1beta1` to `networking.k8s.io/v1` to avoid future breaking change because [deprecation with GKE 1.22](https://cloud.google.com/kubernetes-engine/docs/deprecations/apis-1-22#ingress-v122)
- Remove the `frontend-nodeport.yaml` because it's not anymore a requirement, the default `ClusterIP` one is sufficient/working now.
- To avoid to have another Public IP or any conflict with historical `NodePort` service, do a `kubectl delete svc frontend` in the process too
- The `frontend` service in this folder has also been renamed `frontend-clusterip` to avoid any conflict with existing `frontend` service, because of the `type` field immutable.

Next steps:
- [X] Test on a personal GKE cluster (see comment below with details information about the tests)
- [ ] Review by the team
- [ ] If validated, apply this to [bank-of-anthos.xyz](https://bank-of-anthos.xyz/)


.github/prod-cluster/backend-config.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2021 Google LLC
+# Copyright 2022 Google LLC
```
bourgeoisor, line 1: Copyright headers should reflect the earliest year that the copyright header was instantiated (in this case `2021`). You can either:
- Leave the older year; or
- Append a newer year while keeping the older one (e.g. `2021-2022`)

I prefer the former since it's less maintenance.

mathieu-benoit: Good to know, done.

.github/prod-cluster/frontend-service.yaml
```diff
@@ -14,12 +14,15 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: frontend
+  name: frontend-clusterip
```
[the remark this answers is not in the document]
mathieu-benoit: Good question, I should have documented this better in the description, I just updated the last bullet point to reflect this. Typically, this renaming allows a safe deployment for:
1. This specific deployment process, `kubectl apply -f .github/prod-cluster/frontend-service.yaml` without conflict with existing `frontend` service already in the cluster, because the `type` field is immutable.
2. Regardless this PR, the service by default as `LoadBalancer` (in `~/kubernetes-manifests`) has the same name, so historically deploying this `NodePort` service should have already complained (immutable type field). In OnlineBoutique, the default one is named `service-external`.
Hope that makes sense, with that said, I don't see any issue around that, let me know if you have in mind soemething else.

bourgeoisor (approved):
LGTM, and this doesn't modify any user-facing files so we can merge it at your earliest convenience.

PR #676, opened 2022-03-08
`NetworkPolicy` resources

mathieu-benoit:
Fine granular `NetworkPolicy` resources for Bank Of Anthos


mathieu-benoit:
Let's tackle this later in the future, closing for now.

PR #723, opened 2022-03-31
[Github actions] Add java 17 source profile to push deploy workflow

askmeegs:
Fixes broken push deploy tests on main. 

Importing the custom java env on the runner allows us to override with the required java 17. If we don't import this env, skaffold thinks we're using a different version of JDK that's incompatible with the java 17 version listed in our pom files. 

We're importing this env in our other github workflows (`ci-pr` and `ci-main`)  - example: https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/main/.github/workflows/ci-main.yaml#L117 

— so it just needs to be added to `push and deploy latest`.

### DIFF
--- a/.github/workflows/push-deploy.yml
+++ b/.github/workflows/push-deploy.yml
@@ -26,6 +26,7 @@ jobs:
     - name: Push latest images to GCR
       timeout-minutes: 20
       run: |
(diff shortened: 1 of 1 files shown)


bourgeoisor (approved):
LGTM!

PR #733, opened 2022-04-04
Update JAVA_HOME profile to reflect correct path in `/opt/`

askmeegs:
Directory in the previous step is `/opt/jdk17` so I believe this is what JAVA_HOME should be.


ckim328 (approved):
LGTM!! 🚀

bourgeoisor (approved):
LGTM, good catch!

PR #737, opened 2022-04-05
Substitute version env correctly on release

bourgeoisor:
Previously, the version env was substituted manually in k8s manifests. This change makes sure that we don't miss this step by automating it. Additionally, the dev manifests now has its version statically pinned to `"dev"`.

### DIFF
--- a/dev-kubernetes-manifests/balance-reader.yaml
+++ b/dev-kubernetes-manifests/balance-reader.yaml
@@ -37,7 +37,7 @@ spec:
           readOnly: true
         env:
         - name: VERSION
-          value: "v0.5.3"
+          value: "dev"
         - name: PORT
           value: "8080"
         # toggle Cloud Trace export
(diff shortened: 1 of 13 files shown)


NimJay (approved):
The error in the checks seems unrelated.
```
  1) User can create account
       sees no transactions:
     AssertionError: Timed out retrying: Expected to find element: `#transaction-table`, but never found it.
      at Context.eval (http://34.69.81.143/__cypress/tests?p=cypress/integration/create_spec.js:200:8)
```
Possibly a flaky test (e.g., Cyprus didn't wait until the HTML was fully rendered or sign-up was lagging)?
(On the staging URL, I was able to create an account and even see the HTML element with id `transaction-table`.)

Screen Shot 2022-04-05 at 12 56 11 PM
 
It's worth re-running the check but I'll let you click "re-run" in case you want to quick access the failing error message for now.

In terms of file changes, everything looks good.
I doubt `value: dev` is going to appear somewhere else that we don't want the script to touch. So approved (as long as a re-run results in passing checks). 👍

PR #749, opened 2022-04-18
Added functionality to set location values and change metadata server

arueth:
### Background 
The `frontend` service depends on the metadata server to pull the cluster name and zone. The metadata server is not available to all flavors of Anthos.

### Change Summary
Add the ability to change the metadata server to easily point to a simulator or set the values manually via environment variables.

### Testing Procedure
Tested locally by setting the various permutation of the values in the manifest files. Left commented values in manifest files.


bourgeoisor (approved):
This LGTM

arueth:
Would it be possible to get this merged?

PR #759, opened 2022-04-25
Change default demo password

### LINKED ISSUE #742: Change default demo password
Currently, the default demo password is `password`, which is not necessarily a bad thing, but due to the nature of Chrome's password leak checker, it will always return this when logging in:
image

We should modify the default password to something less common that wouldn't trigger this screen, which may confuse users.

https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/main/dev-kubernetes-manifests/config.yaml#L47
https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/afa662f99c24a8fc88157fb5342570a49eba1445/src/accounts-db/initdb/1-load-testdata.sql#L19-L22

rogerthatdev:
### Background 
The default password as 'password' prompts Chrome to alert the user to a 'data breach'.  

### Change Summary
Changes the default password in the load-testdata.sh script for accounts-db and in the configmap that populates the initial login.
### Testing Procedure
Bring up a new boa cluster; populated username and password should be 'testuser' and 'bankofanthos'. There should be no warning prompt from Chrome.

### Related PRs or Issues 
N/A


src/accounts-db/initdb/1-load-testdata.sh
```diff
@@ -98,9 +98,9 @@ main() {
     fi
   done
 
-  # A password hash + salt for the demo password 'password'
+  # A password hash + salt for the demo password 'bankofanthos'
   # Via Python3:  bycrypt.hashpw('password'.encode('utf-8'), bcrypt.gensalt())
-  DEFAULT_PASSHASH='\x243262243132245273764737474f39777562452f4a786a79444a7263756f386568466b762e634e5262356e6867746b474752584c6634437969346643'
+  DEFAULT_PASSHASH='\x243262243132244c48334f54422e70653274596d6834534b756673727563564b3848774630494d2f34717044746868366e42352e744b575978314e61'
```
bourgeoisor, line 103: Could you also update the hash in here: https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/main/src/accounts-db/initdb/1-load-testdata.sql#L19-L22
(all 4 lines)

rogerthatdev: ack

bourgeoisor (approved):
LGTM

PR #766, opened 2022-04-27
Fixed quotes around favicon

arueth:
### Background 
The favicon href has right double quotes instead of standard double quotes cause it to throw an error in the browser.

### Change Summary
Changed the quotes

### Testing Procedure
Tested locally


bourgeoisor (approved):
Great catch! LGTM.

PR #802, opened 2022-05-25
Initial refactor

aablsk:
### What has happened so far
**skaffold**
- Add skaffold.yaml per service (for maximum flexibility)
- Add skaffold.yaml per deployment unit
- Add profiles for staging, production

**Cloud Build**
- Add reusable cloudbuild.yaml pipeline definition (get cache from GCS, skaffold build, write cache to GCS, create Cloud Deploy release)
- Add cloudbuild-mvnw.yaml pipeline as workaround for an issue when building the java services in the [gcr.io/k8s-skaffold/skaffold](http://gcr.io/k8s-skaffold/skaffold) container ([created a bug](https://github.com/GoogleContainerTools/skaffold/issues/7409))
- Set up one Cloud Build trigger per deploy unit

**Cloud Deploy**
- Add clouddeploy.yaml with staging and production environment; individual delivery pipeline per deploy unit

**kustomize**
- kustomizations per service
- kustomizations per team to easily create reusable components (please note that most of these are for demonstration purposes and not necessarily purposeful)
- kustomizations on top level to easily create reusable components (please note that most of these are for demonstration purposes and not necessarily purposeful; except src/k8s/base/add-public-key-volume which I believe could have actual value)

**GCP reproduction**
- set everything up in GCP to test and showcase it
- automate deployment with pulumi

**make-release.sh**
- switched to skaffold build & render per deploy unit

### TODOs and open questions:

How to deal with inconsistent naming?
- userservice vs contacts
- ledger-db vs ledgerwriter (to hyphen or not to hyphen?)
- balancereader vs balance-reader.yaml (service name in YAML vs YAML file name)

Have not addressed config.yaml
- unsure where this should be located
- I thought about making this a kustomize component but that would prevent us from making it a required module within skaffold - what is best practice here?

How to deal with GCP-project and GCP-resource references in cloudbuild.yaml & clouddeploy.yaml
- solved with pulumi as well as @palladius shell scripts, will probably add terraform as well (as practice for aablsk)

**TODO: ledgermonolith**
- Align team/deploy-unit naming

**make-release.sh**
- Unsure about structure for release manifests: $VERSION/$DEPLOY_UNIT vs $DEPLOY_UNIT/$VERSION hierarchy (I think the second one would be better to give individual teams more control)
- If we want YAML per service we need to run this often or rewrite so it knows about the individual services, which means touching the script every time a new service is added
- How would we integrate make-release in a CI/CD scenario? 
  - Would we have a manually triggered Cloud Build pipeline where the VERSION is explicitly set?
  - Should these be committed to git or would Cloud Deploy be sufficient for versioning?
  - How to name tags?


aablsk:
Created PR with wrong target. Closed immediately.

PR #804, opened 2022-05-26
Bump CPU for userservice for cypress tests 

ckim328:
### Background 
I've been noticing some flaky frontend tests. This is causing PRs to fail on CI tests. Even clicking through the front end on URLs takes a while.
Running tests in my own GCP project, it passed (I only ran it 2 times) 
Only noticed failing tests around the create_spec.js, specifically the User can create account suites in the PRs currently

### Change Summary
Added timeouts to the  cypress create tests.

### Additional Notes
We should remove these waits, and do a deeper investigation why.

### Testing Procedure
Hopefully the CI tests pass

### Related PRs or Issues 
n/a


bourgeoisor (approved):
LGTM

Issue #685, opened 2022-03-22
Bump Google, Flask, and OpenTelemetry dependencies


bourgeoisor:
There are a handful of dependencies that should be bumped that all depend on each other:
```
click
flask
jinja2
itsdangerous
markupsafe
packaging
google-api-core
google-auth
google-cloud-core
google-cloud-trace
google-cloud-monitoring
opentelemetry-sdk
opentelemetry-api
```

Unfortunately, bumping one starts a chain of dependency bumped that ends up in this issue: https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/issues/165#issuecomment-1075372285 (TL;DR: the OpenTelemetry resource detector for GCP has not been updated for `google-auth` in a while and conflicts with other constraints:

```
There are incompatible versions in the resolved dependencies:
  google-auth~=1.22 (from opentelemetry-resourcedetector-gcp==1.1.0a0->-r requirements.in (line 27))
  google-auth<3.0dev,>=1.25.0 (from google-api-core==2.7.1->-r requirements.in (line 11))
  google-auth<3.0dev,>=1.25.0 (from google-cloud-core==2.2.3->-r requirements.in (line 13))
```

Also relevant, the `opentelemetry-tools-google-cloud` package [has been deprecated](https://pypi.org/project/opentelemetry-tools-google-cloud/) in favour of [opentelemetry-resourcedetector-gcp](https://pypi.org/project/opentelemetry-resourcedetector-gcp/) and [opentelemetry-propagator-gcp](https://pypi.org/project/opentelemetry-propagator-gcp/).

bourgeoisor:
Reopening, #775 only updated the frontend deps.

PR #840, opened 2022-06-08
Bump python deps

bourgeoisor:
This bumps more python deps that were leftover from https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/685


mathieu-benoit (approved):
LGTM, all green, all good!

Just curious, why Renovate didn't do/catch these?

bourgeoisor:
Renovate did catch these! See the ignored section as of this comment: https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/599

Screenshot for prosperity since that section will disappear once this merges:
image

But since a lot of these dependencies had to be bumped at the same time (click and flask, or all the google auth / grpcio deps), it had to be a slightly more manual process (which is this PR) to do it all in one go.

PR #883, opened 2022-07-19
Prevent gh actions from running on draft pr

ckim328:
### Background 
On draft PRs, the gh ci runs. Depending on how long a draft is open for, it could use up resources for us.

### Change Summary
Add a type on `ready to review`, and `requested review in` the ci-pr.yaml

### Additional Notes
https://github.com/GoogleCloudPlatform/microservices-demo/pull/926

### Testing Procedure
Open in draft, tests dont run

### Related PRs or Issues 
n/a

### DIFF
--- a/.github/workflows/ci-pr.yaml
+++ b/.github/workflows/ci-pr.yaml
@@ -15,6 +15,7 @@
 name: "Continuous Integration - Pull Request"
 on:
   pull_request:
+    types: [review_requested, ready_for_review]
     branches:
       - main
 jobs:


.github/workflows/ci-pr.yaml
```diff
@@ -15,6 +15,7 @@
 name: "Continuous Integration - Pull Request"
 on:
   pull_request:
+    types: [review_requested, ready_for_review]
```
bourgeoisor, line 18: That definitely covers the initial creation of the PR (or when a draft PR gets converted to a "ready to review" PR) but does it also cover the cases where the PR is rebased, gets new commits, etc? (in other words, will those actions also run this Action?)

ckim328: yes those actions will run

PR #886, opened 2022-08-04
Update wording to match context menu

Mukamik:
The option in the context menu seems to have changed since this doc was written, and "Find command" is no longer in the menu.


ckim328 (approved):
LGTM! Thanks @Mukamik 👍

PR #897, opened 2022-08-23
Updated the DEFAULT_PASSHASH to match the new default password

arueth:
### Background 
The DEFAULT_PASSHASH was not updated in the [populate-accounts-db.yaml](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/main/extras/cloudsql/populate-jobs/populate-accounts-db.yaml) file when the default password was changed.

### Change Summary
Updated the DEFAULT_PASSHASH

### Testing Procedure
Tested via deployment on a GKE standard cluster


bourgeoisor (approved):
LGTM, good catch!

PR #903, opened 2022-08-24
CloudSQL fixes

arueth:
### Background 
While trying to deploy the Bank of Anthos application using CloudSQL I encountered several issues with the populate jobs not completing. It was determined that this was due to the populate jobs having a dependency on the config.yaml manifest

### Change Summary
- Updated to PostgreSQL 14 to match the in-cluster database
- Moved the ConfigMaps for the populate jobs to be before the Job itself.
- Added an apply of the config.yaml file before applying the populate jobs
- Added the runAsNonRoot securityContext to populate-ledger-db.yaml.
- Updated to the Bank of Anthos 0.5.6 version
- Fixed an issue with the timestamp format for one of the populate jobs. The default date utility on Alpine does not support
%N as it is a GNU extension.

### Testing Procedure
Tested in a Qwiklab environment using a GKE Standard cluster.

### DIFF
--- a/extras/cloudsql/README.md
+++ b/extras/cloudsql/README.md
@@ -69,6 +69,7 @@ kubectl create secret -n $NAMESPACE generic cloud-sql-admin \
 This command will also deploy two Kubernetes Jobs, to populate the accounts and ledger dbs with Tables and test data.
 
 ```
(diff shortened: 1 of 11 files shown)


bourgeoisor (approved):
This LGTM!

PR #920, opened 2022-08-31
`openjdk:17-slim`

mathieu-benoit:
`openjdk:17-slim` for `ledgerwriter`, `balancereader`, `transactionhistory` and `ledgermonolith`.

All the Java container images are now in `-slim` like it was already the case for the Python container images.

I also added the `<to>` section (in addition to the `<from>` one already existing) to make sure it's taking the right image, [apparently this field is mandatory](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#to-object).

The images went from the size of 318.2 MB to 296.7 MB.

### DIFF
--- a/src/balancereader/pom.xml
+++ b/src/balancereader/pom.xml
@@ -161,7 +161,7 @@
                 <version>3.2.1</version>
                 <configuration>
                     <from>
-                        <image>openjdk:17</image>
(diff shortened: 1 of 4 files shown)


mathieu-benoit:
Ready for your review, thanks!

bourgeoisor (approved):
LGTM!

PR #946, opened 2022-09-07
Bump locust to 2.12.0

bourgeoisor:
This PR bumps locust to 2.12.0, fixing blocked PR https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/931.


NimJay (approved):
All checks are passing.
Approved!

src/loadgenerator/requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.10
+# This file is autogenerated by pip-compile with python 3.8
```
NimJay, line 2: Thought: I'm going to guess this `3.8` vs `3.10` difference is fine — since it's not a major version difference. :)

bourgeoisor: Yes... I need to bump to 3.10 on my machine, but it won't make a difference here :)

PR #982, opened 2022-09-27
`openjdk:17-slim` --> `openjdk:17-alpine`

mathieu-benoit:
`openjdk:17-slim` --> `openjdk:17-alpine`

Before: 297.2 MB - 51 vulnerabilities
After: 268.3 MB -  16 vulnerabilities

_Note: also did a cleanup with the `<to>` sections not used at all, only `<from>` counts apparently._

### DIFF
--- a/src/balancereader/pom.xml
+++ b/src/balancereader/pom.xml
@@ -163,9 +163,6 @@
                     <from>
                         <image>openjdk:17-alpine</image>
                     </from>
-                    <to>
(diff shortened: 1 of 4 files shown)


mathieu-benoit:
Ready for your review, thanks!

bourgeoisor (approved):
So alpine is more slim than slim? :)
LGTM!

PR #994, opened 2022-10-03
release/v0.5.8

bourgeoisor:
This PR releases a new version of Bank of Anthos.

- Deployed on https://bank-of-anthos.xyz/
- Draft release notes: https://github.com/GoogleCloudPlatform/bank-of-anthos/releases/edit/untagged-0712d7ec82be5ec0c771


mathieu-benoit (approved):
Wondering if this manifests should be updated too https://github.com/GoogleCloudPlatform/bank-of-anthos/tree/main/extras/cloudsql/kubernetes-manifests?

In the future, we need to see why there are these duplicated manifests... maybe with Kustomize in the future we could simplify that...

bourgeoisor:
> maybe with Kustomize in the future we could simplify that...

Yes that's the idea, stay tuned :)

(In the meantime, I've manually commited those changes)

PR #1016, opened 2022-10-14
GKE Autopilot Tutorial: Google Monitoring Dashboard Update

oginskis:
- Added diagrams with metrics coming from frontend's ingress load balancer (request count per second, request size) to the dashboard
- Removed empty diagrams with "rotten" Hibernate related metrics - currently the BOA application does not report any of those


mameshini:
New dashboards for request latency, request count, and response codes work nicely in a lab environment.

bourgeoisor:
Thank you for these changes!

@oginskis instead of creating a new dashboard JSON, could you piggy-back on top of the already existing dashboard.json file? There's a lot of overlap, and I'd prefer keeping everything in one single dashboard. I also notice "GKE Autopilot" mentioned in the dashboard file name -- this dashboard is not strictly restricted to Autopilot and there's nothing inherently different with the dashboard between GKE Standard vs. GKE Autopilot.

TL;DR: modify the dashboard.json instead of creating a new one; and I'll do a pass of reviews then! Thank you!

Amber3232433:
Someone is hacking my devices I don't want any of this on my phone

oginskis:
@bourgeoisor The existing dashboard contains a few charts that depend on things that are not part of the Autopilot tutorial. For example, there is a _Request source to response latency_ chart that depends on metrics coming from Istio. But the deployment of Istio is not in our tutorial (we rely on the Autopilot's built-in ingress controller), therefore that particular Istio-related chart will show nothing if users follow the Autopilot tutorial. Also, users who follow other tutorials (not the Autopilot one), might not see anything in the charts added by us to the dashboard. This could be confusing. So should we merge them or keep them separate?

bourgeoisor:
I think we should try to be as tutorial / external content agnostic, which is why I suggested one dashboard. Could we try that and see how it looks (in the context of your tutorial)? Maybe let's take some screenshots of it and see if it looks too jarring to have empty charts :)

oginskis:
@bourgeoisor I deleted additional dashboard file and added 2 charts to the existing one instead.

Amber3232433:
This is illegal someone is hacking me I don't authorize this

bourgeoisor:
Thank you! It seems like the charts don't show any data if BoA is deployed by itself -- does it only work with the GCE Ingress from https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1017? If so, is there any inherent technical reason preventing metrics from working with the existing frontend LB (https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/main/kubernetes-manifests/frontend.yaml#L139-L150)? If so, what are those constraints?

TheRealSpaceShip:
> Thank you! It seems like the charts don't show any data if BoA is deployed by itself -- does it only work with the GCE Ingress from #1017? If so, is there any inherent technical reason preventing metrics from working with the existing frontend LB (https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/main/kubernetes-manifests/frontend.yaml#L139-L150)? If so, what are those constraints?

In general - yes, it will work with GCE Ingress. This [comment](https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1017#discussion_r1020164194) explains technical details.

bourgeoisor (approved):
LGTM

PR #1017, opened 2022-10-14
GKE Autopilot Tutorial: Horizontal Pod Autoscaler & GCE ingress examples

oginskis:
Added HPA manifest examples:

- scale the userservice deployment based on CPU
- scale the frontend service deployment based on the custom metric received from the loadbalancer

Added Ingress manifest example for the frontend service

### DIFF
--- a/extras/gce-ingress/frontend.yaml
+++ b/extras/gce-ingress/frontend.yaml
@@ -1,0 +1,33 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START gke_boa_kubernetes_manifests_frontend_ingress_frontend]
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: frontend
+  annotations:
+    kubernetes.io/ingress.class: "gce"
+spec:
+  rules:
+  - http:
+      paths:
+      - path: /*
+        pathType: ImplementationSpecific
+        backend:
+          service:
+            name: frontend
+            port:
+              number: 80
+# [END gke_boa_kubernetes_manifests_frontend_ingress_frontend]
--- a/extras/hpa/frontend.yaml
+++ b/extras/hpa/frontend.yaml
@@ -1,0 +1,38 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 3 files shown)


bourgeoisor:
@oginskis instead of creating a new dashboard JSON, could you piggy-back on top of the already existing `dashboard.json` file? There's a lot of overlap, and I'd prefer keeping everything in one single dashboard. I also notice "GKE Autopilot" mentioned in the dashboard file name -- this dashboard is not strictly restricted to Autopilot and there's nothing inherently different with the dashboard between GKE Standard vs. GKE Autopilot.

TL;DR: modify the `dashboard.json` instead of creating a new one; and I'll do a pass of reviews then! Thank you!

extras/gce-ingress/frontend.yaml
```diff
@@ -0,0 +1,33 @@
+# Copyright 2020 Google LLC
```
bourgeoisor, line 1: New files should have <current year>, in this case 2022.

extras/gce-ingress/frontend.yaml
```diff
@@ -0,0 +1,33 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START gke_boa_kubernetes_manifests_frontend_ingress_frontend]
+apiVersion: networking.k8s.io/v1
+kind: Ingress
```
bourgeoisor, line 17: Was there something that wasn't working with the current frontend LB that requires an Ingress? (https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/main/kubernetes-manifests/frontend.yaml#L139-L150)

If so, what are those reasons?

extras/hpa/frontend.yaml
```diff
@@ -0,0 +1,38 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START gke_boa_kubernetes_manifests_frontend_horizontal_pod_autoscaler_frontend]
+apiVersion: autoscaling/v2beta2
+kind: HorizontalPodAutoscaler
```
bourgeoisor, line 17: Could you add a README.md file in this new directory explaining what these HPAs are for, how to deploy it (should be simple -- "go to this directory and kubectl apply..."), to add more context to those ending up here.

extras/gce-ingress/frontend.yaml
bourgeoisor: Same comments on the other new files too.

extras/gce-ingress/frontend.yaml
TheRealSpaceShip: The reason is that we use the external metric `loadbalancing.googleapis.com|https|request_count` which came from LB type HTTP(S). With a Service object type LoadBalancer, it creates Network (TCP and UDP). By using the Ingress object GCE ingress controller creates exactly what we need.

TheRealSpaceShip:
I think we could discard this PR as we have posted a more detailed and advanced one that covers the same topic. https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1085

bourgeoisor:
Closing in favour of https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1085

PR #1041, opened 2022-10-20
`python:3.10.8-slim`

mathieu-benoit:
Like we did in Online Boutique https://github.com/GoogleCloudPlatform/microservices-demo/pull/1168, let's have the Python container base images have now their version as `Ma.Mi.Re` instead of just `Ma.Mi`. As a best practice, this will allow to catch and track CVE vulnerabilities as early as possible.

In this specific case, fixing this: `CVE-2022-40674 | Critical | 9.8 | Yes | expat | OS`.


bourgeoisor (approved):
LGTM

PR #1047, opened 2022-10-23
`eclipse-temurin:17.0.4.1_1-jre-alpine` + `postgres:15.0-alpine`

mathieu-benoit:
Doesn't change anything yet with the current version because `eclipse-temurin:17-jre-alpine` and `eclipse-temurin:17.0.4.1_1-jre-alpine` are the same as of now (same for `postgres:15-alpine` and `postgres:15.0-alpine`). But it's following up on https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1041 and https://github.com/GoogleCloudPlatform/microservices-demo/pull/1204 in order to better track/catch CVEs vulns fix in the future.


bourgeoisor (approved):
LGTM

PR #1054, opened 2022-10-25
GKE Autopilot Tutorial: Resource requirement update for userservice

oginskis:
The Autopilot sets limits to request values (reference: https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-resource-requests#resource-limits) therefore the userservice gets only up to 200 milicores of CPU at maximum on the Autopilot. The userservice generates JWT tokens which are CPU-consuming and currently it takes several seconds for a user to log in. 

I propose to set the CPU requests to 500m to improve the UX. 

@bourgeoisor


bourgeoisor:
Hi @oginskis! Two quick notes:

1) The `kubernetes-manifests/` directory is auto-generated with new releases and should not be manually modified. Instead, could you make your changes in `dev-kubernetes-manifests/`? Those changes are then picked up during the next release.

2) I tested these changes in a GKE Standard default configuration and wasn't able to deploy successfully (not enough vCPU to deploy). Bank of Anthos currently can be deployed in a default GKE Standard cluster without any changes or scaling, and this needs to stay true for the many quickstart flows it supports. I have tried the following which did work, though: bumping the vCPU by 50% from `200m` to `300m`, as well as doubling memory from `64Mi` to `128Mi`. Would that work for you?

Feel free to make those two changes and I'll re-verify this PR

kubernetes-manifests/userservice.yaml
```diff
@@ -80,7 +80,7 @@ spec:
           timeoutSeconds: 10
         resources:
           requests:
-            cpu: 200m
+            cpu: 500m
```
NimJay, line 83: As Olivier mentioned, the `kubernetes-manifests/userservice.yaml` should not be manually updated:
```suggestion
            cpu: 200m
```
Updates to it happen automatically when we create a Bank of Anthos release.
The contents of https://github.com/GoogleCloudPlatform/bank-of-anthos/tree/main/dev-kubernetes-manifests are copied over.
This means that https://github.com/GoogleCloudPlatform/bank-of-anthos/tree/main/dev-kubernetes-manifests is the file that we should make manual updates to.

**Suggestion:** It would be cool if each file inside `kubernetes-manifests/` mentioned this as a warning (inside a comment).

dev-kubernetes-manifests/userservice.yaml
```diff
@@ -80,8 +80,8 @@ spec:
           timeoutSeconds: 10
         resources:
           requests:
-            cpu: 200m
-            memory: 64Mi
+            cpu: 300m
+            memory: 128Mi
```
NimJay, line 84: @bourgeoisor first off, thank you thoroughly testing this pull-request.

**Question:** Have I interpreted [your suggestions](https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1054#issuecomment-1292450452) correctly?

bourgeoisor: Yep, thank you very much, Nim!

bourgeoisor (approved):
LGTM

PR #1097, opened 2022-11-19
Update README.md

chadbr:
iff = if


bourgeoisor (approved):
I believe that the intention was for these to be the shorthand of "if and only if" (https://en.wikipedia.org/wiki/If_and_only_if) and not a typo. That being said, I think the iff is implied anyway, so I'm okay to merge this as-if to improve readability.

PR #1118, opened 2022-11-28
Fix cleanup workflow when namespace is not found

bourgeoisor:
This fixes an occasional issue where the cleanup workflow is attempting to clean-up a namespace that was already cleaned up previously (i.e. it's "not found"), making sure that the workflow doesn't result in a failure if that's the case.

This was seen in a few PRs, more recently: https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1111

image

image


charlieyu1996 (approved):
LGTM

PR #1136, opened 2022-12-02
Unpin ACM version in the Anthos-GKE TF example

### LINKED ISSUE #1135: Terraform example for GKE Anthos cluster setup pins to a specific ACM version
The TF script in the iac/tf-anthos-gke folder specifies a specific ACM version to install in the example. This should be removed so the current release version of ACM is picked automatically.

kaariger:
### Background 
The script was pinning to a specific version as the required features were available in that minimum version. The specified version is now being removed so it picks the current stable version for installation automatically.

### Change Summary
Removed setting the `configmanagement_version` variable in `acm.tf`


bourgeoisor:
The changes in `iac/tf-anthos-gke/acm.tf` look good but do you have any info on that new `terraform` directory? Is that an accidental push?

kaariger:
Yes, good catch. Not sure how that got pushed. I think I merged upstream changes that may be done that. Let me remove that.

bourgeoisor (approved):
LGTM

PR #1156, opened 2022-12-14
Bump opentel to 1.15.0

bourgeoisor:
This PR bumps OpenTel to `1.15.0` and adjacent modules to `0.36b0`.


src/frontend/requirements.txt
```diff
@@ -1,6 +1,6 @@
 #
-# This file is autogenerated by pip-compile with Python 3.11
-# by the following command:
+# This file is autogenerated by pip-compile with python 3.9
+# To update, run:
```
NimJay, line 3: **Thought:** I'm not a Python expert so it probably doesn't matter, but Renovate seems to be using Python 3.11 — and it matches the [Dockerfile's Python version](https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/123dcd50f6fcc4428613f6c99f0ca8813203f4fc/src/frontend/Dockerfile#L16); however, this PR is using Python 3.9. In the past, I've used Docker to generate/output these `requirements.txt` files for Online Boutique (and avoid upgrading Python locally): https://gist.github.com/NimJay/53d95c601dc43721362ab983607f3b7d

NimJay (approved):
Looks awesome! Thanks, Olivier! Approved. Thanks for also generating the `requirements.txt` files.

src/frontend/requirements.txt
bourgeoisor: As long as we're using python 3.7+ it "doesn't matter" what version is used to generate these files (whether 3.9 or 3.11 it'll generate the same format / dependencies) but yes that's a good reminder for me to bump the Python version on my local machine :)

PR #1293, opened 2023-02-22
Updated ledgermonolith README and modified init script to handle multi-instance installations

arueth:
### Background 
The README was not up to date and did not work. You could only use the script to deploy both the service and database to the same instance. We have use cases where we would like to deploy the service and database on separate instances.

### Change Summary
Updated the README and makefile to work properly.
Updated the install-script to allow for installing the service and database on separate instances

### Testing Procedure
Tested using the updated README and [Migrating a monolith VM](https://cloud.google.com/migrate/containers/docs/migrating-monolith-vm-overview-setup)


arueth:
Would also be good to get the `bank-of-anthos-ci` updated.

bourgeoisor (approved):
LGTM!

PR #1338, opened 2023-03-06
Added missing environment variable

arueth:
### Background 
The `initdb/1_create_transactions.sh` script was not running because the `POSTGRES_PASSWORD` was missing from the `--preserve-env` list

### Change Summary
Add missing environment variable to the `--preserve-env` list

### Testing Procedure
Tested locally with my repository


arueth:
It looks like the `POSTGRES_PASSWORD`was added in https://github.com/GoogleCloudPlatform/bank-of-anthos/commit/0629f15b4f0bbfac811a1e8ac1a68d76599e19e0 which had not yet been propigated to the `bank-of-anthos-ci` bucket.

bourgeoisor (approved):
LGTM, nice catch!

PR #1381, opened 2023-03-24
Configure networking for a basic production cluster

TheRealSpaceShip:
This PR introduces configuring networking for a basic production cluster like all in one.

It contains:

* Terraform configuration to create GKE Autopilot cluster
* Terraform configuration to create global IP and DNS Zone
* Modified Kubernetes manifest for `frontend` service. Added custom resources to provision Google-managed SSL certificate, Frontend configuration for Load Balancer and Ingress object to create HTTPS Load Balancer


bourgeoisor:
Instead of Terraform, I would prefer the creation of a simple GKE cluster as well as the IP / DNS configs to be simple `glcoud` commands in the tutorial itself. This reduces complexity and maintainability considerably (e.g., a single `gcloud container clusters create-auto ...` command instead of a 51-lines TF script).

Beyond that, it seems that the only diff to a basic Bank of Anthos deployment are the short `ManagedCertificate` / `Ingress` definitions? It almost seems like you could have the tutorial `kubectl apply` the root `kubernetes-manifests/` directory and then walk through the remaining missing resources manually - no code changes needed on this repo. There's precedent to having users copy-paste a block of code into a file (e.g. https://cloud.google.com/kubernetes-engine/docs/quickstarts/deploy-app-container-image#containerizing_an_app).

I'm in strong favour of closing this PR and adding these short diffs in the tutorial doc itself.

bourgeoisor:
@TheRealSpaceShip I just spoke with our tech writer, this is what we should do:
- Open up a PR against https://github.com/terraform-google-modules/terraform-docs-samples and move the two TF samples there (under a new `container/` directory)
- Open up a PR against https://github.com/GoogleCloudPlatform/kubernetes-engine-samples and move _only the diff between vanilla BoA and what you have in this PR_ (so only the `ManagedCertificate` / `Ingress`, etc, things that changed)

TheRealSpaceShip:
@bourgeoisor got it. Thank you.

PR #1221, opened 2023-01-16
[BREAKING CHANGES] Refactor of CI/CD, kustomize & skaffold

aablsk:
### Background 
Initially the fork that this PR is based on was created to showcase Google Cloud's CICD story encompassing Cloud Build / Cloud Deploy / kustomize / skaffold. Going further we implemented the current bank-of-anthos CI with Cloud Build, integrated pre- and post-deployment tests with skaffold for an attractive devX story from inner to outer loop.

### Change Summary
I'm still working on the change summary and link to a document including rationale for decisions.

### Additional Notes
Will add these while working on the change summary and rationales.

### Testing Procedure
Follow instructions in iac/tf-multienv-cicd-anthos-autopilot/README.md

### Related PRs or Issues 
[Previous PR including terraform infrastructure](https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1006)

PR #1421, opened 2023-04-05
Remove unused region tags

bourgeoisor:
This removes all region tags to clean up the codebase-- This follows [YAGNI principles](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it): if we eventually need any of those tags, we can add exactly the ones needed, instead of trying to catch every possible cases ahead of time (and not needing the vast majority of them).

Removing all 110 region tags only breaks three pages, but those were already broken anyway (since the Skaffold structure changed heavily in https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1221. I have already sent internal changelist #522375716 to fix those pages up.


NimJay (approved):

NimJay:
Thanks for creating Changelist 522392727 to pin the 3 pages to a specific Bank of Anthos commit.
Looks like the Changelist os submitted.
I'm guessing `snippet-bot check` takes some time to pick up on changes:
Screenshot 2023-04-06 at 3 09 00 PM

PR #1425, opened 2023-04-06
Release/v0.6.0

bourgeoisor:
This PR releases Bank of Anthos `v0.6.0`.

This is the first release since the refactoring of the CI/CD pipeline to support Kustomize, Cloud Build, Cloud Deploy, and Artifact Registry. As such, it warrants at minimum a bump of the Y component of SemVer.

Draft release notes: https://github.com/GoogleCloudPlatform/bank-of-anthos/releases/edit/untagged-b4569a5ccb5058050ab7

### About the releasing directory
**Edit:** I've decided to keep the releasing directory as-is (that is, `kubernetes-manifests/`) for now. We can revisit this story later.
 
Historically and still in this PR, the release manifests are stored in the `kubernetes-manifests/` directory. Given that these are not the original / source of truth manifests (those live in `src/` alongside their relevant services) and especially now that they are explicitly rendered (using `skaffold render`), I wonder if perhaps "kubernetes-manifests" is a misleading name. It may make sense to try to align ourselves to Online Boutique who uses a `release/` directory for release manifests ([see here](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main/release)). Of course, this would mean keeping the `kubernetes-manifests/` for a little while longer (essentially having a duplicate) while various tutorials (+ this repo's READMEs and tutorials) get fixed up to the newer directory. That said, if there was ever a time to do that change, this would be it (at the same time as all the CI/CD and rendering changes).

The following may also need to be tweaked / fixed:
- https://github.com/GoogleCloudPlatform/bank-of-anthos/tree/main/iac
- https://github.com/GoogleCloudPlatform/bank-of-anthos/tree/main/extras
- Anywhere else in the repo that mentions `kubernetes-manifests` (can do a search)

### About the releasing manifests
**Edit:** I've decided to manually remove the duplicate k8s entries in for now. I have created a tracking issue: https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/1437

Previously, each services had their own rendered manifest files (alongside the two DB services, the JWT key, and the various ConfigMaps). With the current `skaffold render` setup, a few files goes away:
- `accounts-db.yaml` gets duplicated inside of the contacts and userservice files
- `ledger-db.yaml` gets duplicated inside of the balancereader, ledgerwiter, and transactionhistory files
- `jwt-secret.yaml` and `config.yaml` both get duplicated inside of _every_ services

This is intentionally how the Sakffold renderer _should_ work (if we ask it to render a specific service, it will also render out its direct dependencies) and does still run as long as we make sure that no tutorial or other content gets broken because of it (what if a tutorial is explicitly trying to run `kubectl apply -f config.yaml` for example?) but it does look messy when applying Bank of Anthos on a cluster, getting a handful of `unchanged` from duplicate objects:
```
. . .
secret/jwt-key unchanged
deployment.apps/loadgenerator created
configmap/ledger-db-config unchanged
service/ledger-db unchanged
statefulset.apps/ledger-db unchanged
namespace/bank-of-anthos-development unchanged
serviceaccount/bank-of-anthos unchanged
configmap/demo-data-config unchanged
configmap/environment-config unchanged
configmap/service-api-config unchanged
secret/jwt-key unchanged
service/transactionhistory created
deployment.apps/transactionhistory created
namespace/bank-of-anthos-development unchanged
serviceaccount/bank-of-anthos unchanged
configmap/demo-data-config unchanged
. . .
```
I wonder if it would make sense to try to mimic the previous behaviour as much as possible (no duplicates anywhere, DBs in their own files, and ConfigMaps in their own file). This may unfortunately require manual scripting, since it goes beyond what the Skaffold renderer tries to do. Alternatively, we could render everything in a single file (no duplicate would get generated) but this would mean users couldn't easily pick-and-choose services (and would definitely break some tutorials).

## Testing
I have deployed on a default 3-node GKE Standard cluster:
```
~ kubectl apply -f ./extras/jwt/jwt-secret.yaml
secret/jwt-key created

~ kubectl apply -f ./kubernetes-manifests
configmap/accounts-db-config created
service/accounts-db created
statefulset.apps/accounts-db created
service/balancereader created
deployment.apps/balancereader created
configmap/environment-config created
configmap/service-api-config created
configmap/demo-data-config created
serviceaccount/bank-of-anthos created
service/contacts created
deployment.apps/contacts created
service/frontend created
deployment.apps/frontend created
configmap/ledger-db-config created
service/ledger-db created
statefulset.apps/ledger-db created
service/ledgerwriter created
deployment.apps/ledgerwriter created
deployment.apps/loadgenerator created
service/transactionhistory created
deployment.apps/transactionhistory created
service/userservice created
deployment.apps/userservice created

~ kubectl get pods
NAME                                 READY   STATUS    RESTARTS   AGE
accounts-db-0                        1/1     Running   0          14m
balancereader-69f87fcf6-qpwxl        1/1     Running   0          14m
contacts-557b674495-pxjj8            1/1     Running   0          14m
frontend-64c79c9c57-9zqg4            1/1     Running   0          14m
ledger-db-0                          1/1     Running   0          14m
ledgerwriter-65d4d5b5dc-7l4vs        1/1     Running   0          14m
loadgenerator-5bcb85f657-8v2p8       1/1     Running   0          14m
transactionhistory-8fbc64798-pvrjc   1/1     Running   0          14m
userservice-7bd4fdfbb-frxm5          1/1     Running   0          14m
```


NimJay:
#### Regarding: Open question about the releasing directory

Renaming `kubernetes-manifests/` to `release/` makes sense to me, especially because your plan is cautious:
> "this would mean keeping the kubernetes-manifests/ for a little while longer (essentially having a duplicate)"

#### Regarding: Open question about the releasing manifests

I think it's fine to leave as is (i.e., have duplicated database-related manifests and ConfigMaps).
The "unchanged" `kubetctl apply` output such as
```
serviceaccount/bank-of-anthos unchanged
configmap/demo-data-config unchanged
```
is mostly harmless.
If anything, we could revisit this as an improvement for a later release (e.g., `0.6.1`).

src/ledger/transactionhistory/k8s/base/transactionhistory.yaml
```diff
@@ -44,7 +44,7 @@ spec:
         image: transactionhistory
         env:
         - name: VERSION
-          value: "v0.5.4"
+          value: "dev"
```
NimJay, line 47: Praise: Good catch!

docs/releasing/make-release.sh
```diff
@@ -56,18 +56,20 @@ skaffold build --file-output="artifacts.json" --profile "${PROFILE}" \
...
 # update version in manifests
-find "${REPO_ROOT}/${RELEASE_DIR}" -name '*.yaml' -exec sed -i -e "s'value: \"dev\"'value: \"${NEW_VERSION}\"'g" {} \;
-rm "${REPO_ROOT}/${RELEASE_DIR}/*-e"
+find "${REPO_ROOT}/${RELEASE_DIR}" -name '*.yaml' -exec sed -i -e "s'value: dev'value: ${NEW_VERSION}'g" {} \;
+rm ${REPO_ROOT}/${RELEASE_DIR}/*-e
 
 # update version in terraform scripts
 sed -i -e "s@sync_branch  = .*@sync_branch  = \"${NEW_VERSION}\"@g" ${REPO_ROOT}/iac/tf-anthos-gke/terraform.tfvars
-rm "${REPO_ROOT}/iac/tf-anthos-gke/terraform.tfvars-e"
+rm ${REPO_ROOT}/iac/tf-anthos-gke/terraform.tfvars-e
```
NimJay, line 72: **Question:** Are we removing this to ensure that it doesn't get committed? I wonder if it makes more sense to use `.gitignore`.

docs/releasing/header.txt
```diff
@@ -1,19 +1,13 @@
...
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: bank-of-anthos-development
-  labels:
-    istio-injection: enabled # enable ASM sidecar injection on namespace
```
NimJay, line 19: **Thought:**
Regarding folder: `/iac/acm-multienv-cicd-anthos-autopilot/overlays`
I see that we're removing `development/namespace.yaml`.
Note that we're also removing the `istio-injection: enabled` label.
So this means that the development manifests will use the `default` namespace. That's fine.
But the development will behave differently (from production and staging) when using Istio.
Namely, users will have to manually label the namespace with `istio-injection: enabled` to have Istio Sidecars auto-injected.

NimJay: Not a big issue. Feel free to ignore.

docs/releasing/header.txt
```diff
@@ -1,19 +1,13 @@
-# Copyright 2022 Google LLC
+# Copyright 2023 Google LLC
```
NimJay, line 1: **Thought:**
I guess this means that the release manifests will always use the current year?
That is, if we introduce a new file inside `kubernetes-manifest` in 2024, we would have to bump `2023` here to `2024`.
Fine by me.

NimJay (approved):
Thank you again, @bourgeoisor, for the walkthrough of this pull-request yesterday.
For the open questions, I would consider opening a new GitHub issue just for tracking purposes — and work on those improvements incrementally.
I think it's important to not stuff too many changes into this pull-request.

Excellent work!
I trust your testing of the `kubernetes-manifests/`. I have not tested them myself.
Approved.

docs/releasing/make-release.sh
bourgeoisor: Yes that's residual from `sed`. Sure, it _could_ be added to the `.gitignore` but my point here is that I don't want them lingering on my machine in the first place, so I'm cleaning them up here :)

docs/releasing/header.txt
bourgeoisor: "So this means that the development manifests will use the default namespace"

Yes. It already does, this `bank-of-anthos-development` namespace actually just sat empty. I think it's residual from an earlier thought that we'd use a specific namespace for development work, but that wasn't compatible with the idea of multiple deployments in the same cluster (e.g. PR clusters) so that idea was scrapped.

"Namely, users will have to manually label the default namespace with istio-injection: enabled to have Istio"

That's fine, neither raw istio or ASM are deployed in the development profile.

docs/releasing/header.txt
bourgeoisor: My take here is that technically, every file in `kubernetes-manifests/` is _always_ a new file, because it's newly generated / rendered from scratch at release time, as opposed to other more static files (like from `src/`) which wouldn't have their header changing and would keep the "earlier" date.

kubernetes-manifests/frontend.yaml
```diff
@@ -1,147 +1,198 @@
...
+  BALANCES_API_ADDR: balancereader:8080
+  CONTACTS_API_ADDR: contacts:8080
+  HISTORY_API_ADDR: transactionhistory:8080
+  TRANSACTIONS_API_ADDR: ledgerwriter:8080
+  USERSERVICE_API_ADDR: userservice:8080
+kind: ConfigMap
+metadata:
+  name: service-api-config
+---
+apiVersion: v1
```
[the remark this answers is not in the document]
bourgeoisor: Yep! It simplifies a lot :)

PR #1507, opened 2023-05-12
Update default sync_branch var to point to branch

LisaFC:
### Background 
Sync branch was actually pointing to a tag name rather than a branch name, which meant it wasn't picking up all the changes in the release branch (including a changed directory name).

### Change Summary
Changed the default value for this variable


bourgeoisor (approved):
LGTM!

PR #1578, opened 2023-06-12
Add new wait resource

LisaFC:
Add wait that we can use in acm.tf. This helps ensure that Bank of Anthos doesn't start up until the Istio injection label is applied so we don't need to restart to get proxies.

### Background 
Before this PR, we saw inconsistent behaviour when applying Bank of Anthos - sometimes the ASM proxies were up for all services when the apply was complete, sometimes for only some services, and sometimes for no services. Users needed to restart all pods to get the desired ASM behaviour.

### Change Summary
- Added new wait resource
- Added dependency on the wait resource in acm.tf
- Removed enabling metrics SA in acm.tf as this doesn't seem to be required any more and causes issues with adding dependencies

### DIFF
--- a/iac/tf-anthos-gke/acm.tf
+++ b/iac/tf-anthos-gke/acm.tf
@@ -29,5 +29,7 @@ module "acm" {
 
   secret_type = "none"
     
-  create_metrics_gcp_sa = true
+  create_metrics_gcp_sa = false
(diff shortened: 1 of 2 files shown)


bourgeoisor (approved):
LGTM!

PR #1740, opened 2023-08-21
cleanup: Use e2 series machine instead of n1 for the ledgermonolith VM

macribas:
This repo is used in the Anthos labs in Qwiklabs. We are asking every lab developer to stop using N1 series machines and use only E2 series machines, which are cheaper to run and less prone to resource starvation issues.

## Before
image

## After
image

### DIFF
--- a/src/ledgermonolith/scripts/deploy-monolith.sh
+++ b/src/ledgermonolith/scripts/deploy-monolith.sh
@@ -64,7 +64,7 @@ gcloud compute instances create ledgermonolith-service \
     --subnet ${SUBNET=default} \
     --image-family=debian-10 \
     --image-project=debian-cloud \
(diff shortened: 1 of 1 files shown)


bourgeoisor:
@macribas have you tested the lab / deploying the monolith with this changed machine type and that it works as expected?

macribas:
Yes, that lab is working fine with that configuration. However I don't
know where else this script is used.

bourgeoisor:
It's also used for this tutorial: https://cloud.google.com/migrate/containers/docs/migrating-monolith-vm-overview-setup

But if the lab is working as expected, I'm confident in merging this. Thanks Marcel!

bourgeoisor (approved):

PR #1802, opened 2023-09-20
Remove preview skaffold version from Cloud Build / Deploy step

bourgeoisor:
This PR removes the experimental Skaffold version flag from the Cloud Deploy release creation. This flag used to be necessary when it was added since we were relying on Skaffold changes that were not included in the default Cloud Deploy release, but nowadays Cloud Deploy uses Skaffold 2.6 by default which is more than sufficient for our needs.


minherz (approved):
lgtm

PR #1898, opened 2023-10-17
feat(renovate): group PRs by language and schedule weekly

muncus:
This new renovate config will run weekly, and group PRs based on the "language"
(actually the renovate manager).

This should reduce the number of PRs and PR conflicts significantly, down to
~5/wk instead of ~15. And eliminate issues with tightly-coupled dependencies.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,15 +1,22 @@
 {
   "extends": [
-    "config:base"
+    "config:recommended",
+    ":approveMajorUpdates",
   ],
+  // Run once a week, outside typical office hours.
+  // TZ is UTC unless specified.
+  "schedule": [ "after 10pm on Tuesday" ],
   "pip-compile": {
     "enabled": true,
     "fileMatch": ["(^|/)requirements\\.in$"]
   },
+  // pip-compile handles requirements.txt updates.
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -0,0 +1,102 @@
...
+     "labels":[
+       "type:security"
+     ],
+     "stabilityDays":0
+  },
+  "labels": ["dependencies"],
+  "kubernetes": {
+    "fileMatch": ["\\.yaml$"],
+    "ignorePaths": [
+      "release/**",
```
bourgeoisor, line 31:
```suggestion
      "kubernetes-manifests/**",
```

.github/renovate.json5
```diff
@@ -0,0 +1,102 @@
...
+      "kustomize/base/**"
```
bourgeoisor, line 32: This is not a directory that exists; there are multiple bases within `src/` (though I don't think they need to be skipped necessarily)

muncus: oops, this is what i get for copying the file from Online Boutique 🤦🏼 
Thanks for noticing!

bourgeoisor (approved):

PR #1957, opened 2023-12-12
Bump Python & Docker deps

bourgeoisor:
## The problem
Bulk Renovate PRs for the following were all failing at the e2e test for various cryptic reasons:
- https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1951
- https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1939
- https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1938

So I decided to merge them all in the same branch (this one) to deal with them in one go.

## The culprits
Turns out there were two different issues, which all three PRs were triggering (either one of the issues, or both):

a) https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1957/commits/386bc976e55fa8663e6a81af75efe0337b387d5b (Werkzeug v3 -- which is required for Flask v3 but which was also getting bumped in the regular Python batch -- fixed an issue where `set_cookie()` was accepting bytes when it should've only accepted strings. We were doing the latter and was caught in the middle. Making sure we only report a string fixes these failures.)
image

b) https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1957/commits/c54986e154c3f6ab972fd9287f74f18a030b20ef (Python 3.12 deprecated a behaviour where `random.randint()` could accept floats. It now only accepts ints. Making sure we're only using ints fixes these failures.)
image


NimJay (approved):
Looks good! Thanks for fixing this.

src/accounts/userservice/db.py
```diff
@@ -70,7 +70,7 @@ def generate_accountid(self):
         accountid = None
         with self.engine.connect() as conn:
             while accountid is None:
-                accountid = str(random.randint(1e9, (1e10 - 1)))
+                accountid = str(random.randint(1_000_000_000, (10_000_000_000 - 1)))
```
NimJay, line 73: Today I learned you can define numbers like `10_000` (using underscores) in Python. Very cool feature.

bourgeoisor: Yep! Technically not required (I could've just put `1000000000`) but it helps with readability.

The fix wasn't the underscore, but rather going from `1e9` (which is a `float`) to an `int`.

PR #1978, opened 2024-01-10
Change live demo domain to cymbal.dev

bourgeoisor:
This PR changes the live demo URL from `bank-of-anthos.xyz` to `cymbal-bank.fsi.cymbal.dev`.


NimJay (approved):
Approved! Thanks for doing this.
Waiting until tomorrow (as your suggested) before merging sounds good to me.

PR #1990, opened 2024-01-19
Database job and Dockerfile cleanup

arueth:
### Change Summary
Updated the jobs to use the same container image as the database
Cleaned up the accounts-db Dockerfile
Cleaned up the ledger-db Dockerfile

### Testing Procedure
Tested in my project


bourgeoisor (approved):
LGTM

PR #2042, opened 2024-03-04
removing reference to obsolete sample deployment marketplace app

### LINKED ISSUE #2043: should remove reference to obsoleted click-to-deploy marketplace sample app
We're trying to clean up the Anthos Sample Deployments and came across a tutorial link in this project's root README file. This ticket is to just remove it; we can subsequently link to terraform content for the same info in another pass, critical path here is to just remove the reference so we can take down the marketplace app safely.

linde:
### Background 
noticed the sample has a link to a marketplace app we're trying to remove. just wanted to delete the link.

### Change Summary
just a link removal from the root readme.

### Additional Notes
n/a

### Testing Procedure
however the main readme is checked, this is not a major change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,6 @@ The following button opens up an interactive tutorial showing how to deploy Bank
 - [Troubleshooting](/docs/troubleshooting.md) to learn how to resolve common problems.
 
 ## Demos featuring Bank of Anthos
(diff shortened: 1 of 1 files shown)


linde:
hey, I am wrong -- this URL points to the update tutorial that uses terraform and is supported. nevermind, sorry for any bother!

PR #2044, opened 2024-03-05
Disabled spring.cloud.gcp.trace to stop ADC lookup

arueth:
### Background 
Even with `spring.sleuth.enabled` set to `false`, zipkin2 tried to lookup the ADC credentials. This causes the application to crash on some environments. 

### Change Summary
Added `spring.cloud.gcp.trace.enabled`  to be set via the `ENABLE_TRACING` environment variable as well to fully disable the trace functionality in the apps.

### Testing Procedure
Tested on my personal machine using Docker Desktop.


bourgeoisor (approved):
LGTM!

PR #2166, opened 2024-07-24
added the user and set the necessary permissions

### LINKED ISSUE #517: Components should not require root to run
In clusters with PodSecurityPolicies active and restrictions on which images can run as root, it's hard to deploy this demo application. The DB components fail to start when trying to start as non root (using the postgres user (id 70) in the securityContext). 
The app components requiring the JWT secret also fail to start as the key is mounted and expected in the /root directory.

rock-007:
### Background 
The DB components fail to start when trying to start as non root user.

### Change Summary
I have added group and user as postgres and set the permissions as raised by ["mathieu-benoit"](https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/517) in the ticket ref:517

### Related PRs or Issues 
#517


rock-007:
I have signed the CLA, not sure why it still says missing CLA here?

PR #855, opened 2022-06-21
Unprivileged containers

mathieu-benoit:

PR #2167, opened 2024-07-25
added the user and set the necessary permissions

### LINKED ISSUE #517: Components should not require root to run

rock-007:
### Background 
The DB components fail to start when trying to start as non root user.

### Change Summary
I have added group and user as postgres and set the permissions as raised by ["mathieu-benoit"](https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/517) in the ticket ref:517

### Related PRs or Issues 
#517


bourgeoisor:
Hi @rock-007, have you tested this locally? The Docker image doesn't build.

```
docker build failure: The command '/bin/sh -c addgroup -S postgres && adduser -S postgres -G postgres' returned a non-zero code
```

src/ledger/ledger-db/Dockerfile
```diff
@@ -16,6 +16,16 @@ FROM postgres:16.3-alpine@sha256:de3d7b6e4b5b3fe899e997579d6dfe95a99539d154abe03
 # Need to get coreutils to get the date bash function working properly:
 RUN apk add --no-cache coreutils && rm -rf /var/cache/apk/*
 
+# Create a user and group with the same UID and GID as the postgress
+RUN if ! getent group postgres > /dev/null; then addgroup -S postgres; fi && \
+    if ! getent passwd postgres > /dev/null; then adduser -S postgres -G postgres; fi
```
bourgeoisor, line 21: Container images should have deterministic build steps. Either these groups / bindings exist, or they don't. We should need conditional checks.

rock-007: removed the conditional check.

bourgeoisor:
This PR would need changes to the k8s YAML too, e.g. https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/855/files

rock-007:
> This PR would need changes to the k8s YAML too, e.g. https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/855/files

Done.

kubernetes-manifests/ledger-db.yaml
```diff
@@ -103,6 +103,11 @@ spec:
             - mountPath: /var/lib/postgresql/data
               name: postgresdb
               subPath: postgres
+          securityContext:
```
bourgeoisor, line 106: That's a start but it's also missing security contexts to prevent all access to root:

```
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
              - all
          privileged: false
          readOnlyRootFilesystem: true
```

rock-007: I have added these now to the `ledger-db.yaml` file.

src/ledger/ledger-db/Dockerfile
```diff
@@ -16,6 +16,12 @@ FROM postgres:16.3-alpine@sha256:de3d7b6e4b5b3fe899e997579d6dfe95a99539d154abe03
 # Need to get coreutils to get the date bash function working properly:
 RUN apk add --no-cache coreutils && rm -rf /var/cache/apk/*
 
+# Change ownership of the necessary directories
+RUN chown -R postgres:postgres /var/lib/postgresql /var/run/postgresql
+
+# Set thte correct permissions
+RUN chmod -R 0700 /var/lib/postgresql/data && chmod -R 0755 /var/run/postgresql
```
bourgeoisor, line 23: @mathieu-benoit do you think that's all that was needed? Seems too good to be true but maybe I'm overthinking it

STARLING077 (approved):

STARLING077 (approved):

PR #2201, opened 2024-08-30
fix: Pin Google Terraform providers to known good version

### LINKED ISSUE #2200: Config Sync not installed/enabled for tf-multienv-cicd-anthos-autopilot clusters
### Describe the bug 
For the `tf-multienv-cicd-anthos-autopilot` deployment, clusters fail to synchronize because Config Sync is not enabled.

### To Reproduce 
When following [the documentation](tf-multienv-cicd-anthos-autopilot), checking the Config dashboard (or running `gcloud beta container hub config-management status`) will show that clusters never synchronize.

### Additional context 
This is caused by a recent [change](https://github.com/hashicorp/terraform-provider-google/pull/19016) to the Google Terraform providers, and can be fixed by pinning to a known good version. I have a PR incoming for this.

I will follow up with the provider authors to understand more about this change, as I am not sure this difference in behaviour is intentional.

henrybell:
### Background 
With recent versions of the Google Terraform providers, Config Sync was not being enabled in clusters for the `tf-multienv-cicd-anthos-autopilot` deployment, causing clusters to fail to sync.

### Change Summary
Pin the providers to a tested and known good version, pending further investigation into the provider changes.

###
Fixes #2200.

### DIFF
--- a/iac/tf-multienv-cicd-anthos-autopilot/main.tf
+++ b/iac/tf-multienv-cicd-anthos-autopilot/main.tf
@@ -19,10 +19,12 @@ terraform {
   }
   required_providers {
     google = {
-      source = "hashicorp/google"
+      source  = "hashicorp/google"
+      version = "~> 5.40.0"
     }
     google-beta = {
-      source = "hashicorp/google-beta"
+      source  = "hashicorp/google-beta"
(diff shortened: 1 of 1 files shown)


bourgeoisor (approved):
LGTM

PR #2216, opened 2024-09-09
Remove gcr.io references in the docs

bourgeoisor:
This PR removes references to `gcr.io` in the docs.

Release images have all moved to `us-central1-docker.pkg.dev` a long time ago, and that's what the instructions should use too.

The only references left are those of older images in `extras/`, but those images remain accessible. When those extras are updated with newer images they'll use the `us-central1-docker.pkg.dev` tags.


NimJay (approved):
Looks good to me.

